PR #120, opened 2020-05-06
73 table column sort

dougnoel:


dougnoel:
I'm ok with the state of the code on AlphaNumComparator. It's someone else's code and I'm not going to refactor it now.

PR #210, opened 2021-04-29
Added script to support selenium grid

mdsapon:
To run on test on Grid, provide "runOnGrid" to true or false, if true then provide "gridUrl" , "browser" and "browserVersion" in the yml file or commandline.

Forked from the master branch: https://github.com/mdsapon/sentinel/tree/ImplementGrid

### DIFF
--- a/src/main/java/com/dougnoel/sentinel/webdrivers/WebDriverFactory.java
+++ b/src/main/java/com/dougnoel/sentinel/webdrivers/WebDriverFactory.java
@@ -1,12 +1,14 @@
 package com.dougnoel.sentinel.webdrivers;
 
+import java.net.URL;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.openqa.selenium.WebDriver;
 import org.openqa.selenium.ie.InternetExplorerDriver;
 import org.openqa.selenium.ie.InternetExplorerOptions;
+import org.openqa.selenium.remote.DesiredCapabilities;
+import org.openqa.selenium.remote.RemoteWebDriver;
 import org.openqa.selenium.safari.SafariDriver;
-
 import com.dougnoel.sentinel.configurations.ConfigurationManager;
 import com.dougnoel.sentinel.exceptions.WebDriverNotExecutableException;
 import com.dougnoel.sentinel.exceptions.WebDriverException;
@@ -92,32 +94,57 @@ public class WebDriverFactory {
         String downloadDirectory = ConfigurationManager.getOptionalProperty("download");
         if (downloadDirectory != null)
             DownloadManager.setDownloadDirectory(downloadDirectory);
-
-        String browser = ConfigurationManager.getBrowserName();
-
-        // Initialize the driver object based on the browser and operating system (OS).
-        // Throw an error if the value isn't found.   	
-    	switch (browser) {
-        case "chrome":
-        	String headless = ConfigurationManager.getOptionalProperty("headless");
-        	if (headless == null || headless.equalsIgnoreCase("false"))
-        		driver = ChromeDriverFactory.createChromeDriver();
-        	else
-        		driver = ChromeDriverFactory.createHeadlessChromeDriver();
-            break;
-        case "firefox":
-        	driver = FirefoxDriverFactory.createFirefoxDriver();
-            break;
-        case "internetexplorer":
-        	driver = createInternetExplorerDriver();
-            break;
-        case "safari":
-        	driver = createSafariDriver();
-            break;
-        default:
-            throw new WebDriverException(SentinelStringUtils.format("Invalid browser type '{}' passed to WebDriverFactory. Could not resolve the reference. Check your spelling. Refer to the Javadoc for valid options.", browser));
+        
+        String runOnGrid = ConfigurationManager.getOptionalProperty("runOnGrid");
+    	String browser = ConfigurationManager.getBrowserName();
+    	
+        if (runOnGrid != null && runOnGrid.equals("true")) {
+        	String gridUrl = ConfigurationManager.getOptionalProperty("gridUrl");
+        	String browserVersion = ConfigurationManager.getOptionalProperty("browserVersion");
+        	
+        	//use full browser version from grid in yml or command line e.g: 88.0.4324.96
+        	driver = instantiateRemoteWebDriver(browser, browserVersion, gridUrl);
+        }else {
+            // Initialize the driver object based on the browser and operating system (OS).
+            // Throw an error if the value isn't found.   	
+        	switch (browser) {
+            case "chrome":
+            	String headless = ConfigurationManager.getOptionalProperty("headless");
+            	if (headless == null || headless.equalsIgnoreCase("false"))
+            		driver = ChromeDriverFactory.createChromeDriver();
+            	else
+            		driver = ChromeDriverFactory.createHeadlessChromeDriver();
+                break;
+            case "firefox":
+            	driver = FirefoxDriverFactory.createFirefoxDriver();
+                break;
+            case "internetexplorer":
+            	driver = createInternetExplorerDriver();
+                break;
+            case "safari":
+            	driver = createSafariDriver();
+                break;
+            default:
+                throw new WebDriverException(SentinelStringUtils.format("Invalid browser type '{}' passed to WebDriverFactory. Could not resolve the reference. Check your spelling. Refer to the Javadoc for valid options.", browser));
+            }
         }
 
+        
+
+        return driver;
+    }
+    
+    private static WebDriver instantiateRemoteWebDriver(String browser, String browserVersion, String gridUrl) {
+
+    	DesiredCapabilities capability = new DesiredCapabilities();
+    	capability.setCapability("browserName", browser);
+    	capability.setCapability("browserVersion", browserVersion);
+    	
+        try {
+            driver = new RemoteWebDriver(new URL(gridUrl), capability);
+        } catch (Exception e) {
+            e.printStackTrace();
+        }
         return driver;
     }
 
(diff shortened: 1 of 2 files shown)


src/main/java/com/dougnoel/sentinel/webdrivers/WebDriverFactory.java
```diff
@@ -92,32 +94,57 @@ public static WebDriver instantiateWebDriver() {
         String downloadDirectory = ConfigurationManager.getOptionalProperty("download");
         if (downloadDirectory != null)
             DownloadManager.setDownloadDirectory(downloadDirectory);
+        
+        String runOnGrid = ConfigurationManager.getOptionalProperty("runOnGrid");
```
dougnoel, line 98: After looking at the information needed, this variable is unnecessary. Just evaluate on the  `gridUrl ` value being set or not here and eliminate the use of the `turnGridOn` value.

src/main/java/com/dougnoel/sentinel/webdrivers/WebDriverFactory.java
```diff
@@ -92,32 +94,57 @@ public static WebDriver instantiateWebDriver() {
...
+    	String browser = ConfigurationManager.getBrowserName();
+    	
+        if (runOnGrid != null && runOnGrid.equals("true")) {
+        	String gridUrl = ConfigurationManager.getOptionalProperty("gridUrl");
+        	String browserVersion = ConfigurationManager.getOptionalProperty("browserVersion");
+        	
+        	//use full browser version from grid in yml or command line e.g: 88.0.4324.96
+        	driver = instantiateRemoteWebDriver(browser, browserVersion, gridUrl);
```
dougnoel, line 106: Is there functionality for the  passing of `latest` or something similar in grid so that we can have a default?

src/main/java/com/dougnoel/sentinel/webdrivers/WebDriverFactory.java
```diff
@@ -92,32 +94,57 @@ public static WebDriver instantiateWebDriver() {
...
             DownloadManager.setDownloadDirectory(downloadDirectory);
+        
+        String runOnGrid = ConfigurationManager.getOptionalProperty("runOnGrid");
+    	String browser = ConfigurationManager.getBrowserName();
+    	
+        if (runOnGrid != null && runOnGrid.equals("true")) {
+        	String gridUrl = ConfigurationManager.getOptionalProperty("gridUrl");
+        	String browserVersion = ConfigurationManager.getOptionalProperty("browserVersion");
+        	
+        	//use full browser version from grid in yml or command line e.g: 88.0.4324.96
```
dougnoel, line 105: This comment is not needed here. It needs to be in the readme. There should be a new FAQ entry, `How do I use Selenium Grid?`

src/main/java/com/dougnoel/sentinel/webdrivers/WebDriverFactory.java
```diff
@@ -92,32 +94,57 @@ public static WebDriver instantiateWebDriver() {
         String downloadDirectory = ConfigurationManager.getOptionalProperty("download");
         if (downloadDirectory != null)
             DownloadManager.setDownloadDirectory(downloadDirectory);
+        
+        String runOnGrid = ConfigurationManager.getOptionalProperty("runOnGrid");
+    	String browser = ConfigurationManager.getBrowserName();
+    	
+        if (runOnGrid != null && runOnGrid.equals("true")) {
+        	String gridUrl = ConfigurationManager.getOptionalProperty("gridUrl");
+        	String browserVersion = ConfigurationManager.getOptionalProperty("browserVersion");
```
dougnoel, line 103: This implementation assumes only one browser version will be used for the whole execution. The point of grid is to support multiple executions. How does this implementation take that into account?

src/main/java/com/dougnoel/sentinel/webdrivers/WebDriverFactory.java
```diff
@@ -92,32 +94,57 @@ public static WebDriver instantiateWebDriver() {
...
+        	
+        	//use full browser version from grid in yml or command line e.g: 88.0.4324.96
+        	driver = instantiateRemoteWebDriver(browser, browserVersion, gridUrl);
+        }else {
```
dougnoel, line 107: Instead of an if/else here, follow the pattern above for Saucelabs and create a GridDriverFactory class that returns a driver object and end execution by returning that driver.

src/main/java/com/dougnoel/sentinel/webdrivers/WebDriverFactory.java
```diff
@@ -92,32 +94,57 @@ public static WebDriver instantiateWebDriver() {
...
-        case "safari":
-        	driver = createSafariDriver();
-            break;
-        default:
-            throw new WebDriverException(SentinelStringUtils.format("Invalid browser type '{}' passed to WebDriverFactory. Could not resolve the reference. Check your spelling. Refer to the Javadoc for valid options.", browser));
-        }
+        return driver;
+    }
+    
+    private static WebDriver instantiateRemoteWebDriver(String browser, String browserVersion, String gridUrl) {
```
dougnoel, line 137: Put this in the GridWebDriverFactory

src/main/java/com/dougnoel/sentinel/webdrivers/WebDriverFactory.java
```diff
@@ -92,32 +94,57 @@ public static WebDriver instantiateWebDriver() {
...
 
+    	DesiredCapabilities capability = new DesiredCapabilities();
+    	capability.setCapability("browserName", browser);
+    	capability.setCapability("browserVersion", browserVersion);
+    	
+        try {
+            driver = new RemoteWebDriver(new URL(gridUrl), capability);
+        } catch (Exception e) {
```
dougnoel, line 145: We should never catch and dump an error. And we should always be as specific as possible with error handling. What exceptions are we expecting here? Your IDE will tell you if you remove the try/catch block. Catch it and throw a well-formatted WebDriverException() to give the end user as much possible information.

src/main/java/com/dougnoel/sentinel/webdrivers/WebDriverFactory.java
```diff
@@ -92,32 +94,57 @@ public static WebDriver instantiateWebDriver() {
...
+            e.printStackTrace();
```
dougnoel, line 146: If you need to write something to console, use the log4j2 logger. You can see examples of it in other files, or I can help you out.

dougnoel (changes requested):

src/main/java/com/dougnoel/sentinel/webdrivers/WebDriverFactory.java
mdsapon: This parallel execution instructions will be taken care by tesng.xml which will send info to grid. so no issues here.

src/main/java/com/dougnoel/sentinel/webdrivers/WebDriverFactory.java
mdsapon: If we don't use try catch then you have to throw malformedurlexception, should I use that and remove try catch ?

src/main/java/com/dougnoel/sentinel/webdrivers/WebDriverFactory.java
mdsapon: let me know about the above comment, based on that I can use log4j or remove try catch

src/main/java/com/dougnoel/sentinel/webdrivers/WebDriverFactory.java
mdsapon: ok, will do

src/main/java/com/dougnoel/sentinel/webdrivers/WebDriverFactory.java
mdsapon: ok, will do

src/main/java/com/dougnoel/sentinel/webdrivers/WebDriverFactory.java
mdsapon: will remove this comment

src/main/java/com/dougnoel/sentinel/webdrivers/WebDriverFactory.java
mdsapon: few things we will do in next sprint: 
1. when we will implement webdrivermanager, then we will handle latest browser version at the same time
2. will work with tony(devops) to use short browser versions e.g 88, 89, 90 etc
3. will think about it how to interpret latest when there will be multiple browser versions

src/main/java/com/dougnoel/sentinel/webdrivers/WebDriverFactory.java
mdsapon: ok, make sense. will do that

src/main/java/com/dougnoel/sentinel/webdrivers/WebDriverFactory.java
dougnoel: Ok, make a ticket to keep track of this work now then I'll resolve this comment.

src/main/java/com/dougnoel/sentinel/webdrivers/WebDriverFactory.java
dougnoel: Catch the MalformedURLException and wrap it in one of ours and throw it. Take a look at lines 25-29 of the SaucelabsDriverFactory to see how I handled it. Not that I'm throwing a sentinel Exception there even though it looks the same.

And yes, we absolutely should be throwing exceptions like this because without a driver we cannot continue.

src/main/java/com/dougnoel/sentinel/webdrivers/WebDriverFactory.java
dougnoel: You don't need to log anything, just wrap and throw the exception.

PR #228, opened 2021-05-25
Fixed new window/tab issue and added unit tests

paglajewel:


src/test/java/com/dougnoel/sentinel/pages/PageManagerTests.java
```diff
@@ -0,0 +1,47 @@
...
+	@Test(expected = com.dougnoel.sentinel.exceptions.NoSuchWindowException.class)
+	public void getWindowHandleInSameWindow() {
+		PageManager.switchToNewWindow();
+	}
+	
+	@Test
+	public void getWindowHandleInNewWindow() {
+		PageManager.openPage(baseUrl);
+		pageElement.getElement("open_new_window").click();
+		PageManager.switchToNewWindow();
```
dougnoel, line 38: There's no assert here telling us what we expect.

src/test/java/com/dougnoel/sentinel/pages/PageManagerTests.java
```diff
@@ -0,0 +1,47 @@
...
+	}
+	
+	@Test
+	public void getWindowHandleInNewTab() {
+		PageManager.openPage(baseUrl);
+		pageElement.getElement("open_new_tab").click();
+		PageManager.switchToNewWindow();
```
dougnoel, line 45: There's no assert here telling us what we expect.

src/test/java/com/dougnoel/sentinel/pages/SampleTestPage.yml
```diff
@@ -0,0 +1,9 @@
+urls:
```
dougnoel, line 1: If we are going to use this page, it should be in a package named com.encodedna to follow the naming convention of newer pages. This is cleanup that still needs to be done on some of the older page objects.

Separately, I don't know what the permissions are for using this project are. We would be better off leveraging the internet herokuapp domain as it's intended for that use. See if this page will get us what we need: https://the-internet.herokuapp.com/windows

dougnoel (changes requested):

dougnoel:
You will need to merge from master.

src/test/java/com/dougnoel/sentinel/pages/SampleTestPage.yml
paglajewel: Herokuapp doesn't have pop up window, it only has new tab

src/test/java/com/dougnoel/sentinel/pages/PageManagerTests.java
paglajewel: done

src/test/java/com/dougnoel/sentinel/pages/PageManagerTests.java
paglajewel: done

src/test/java/com/dougnoel/sentinel/pages/PageManagerTests.java
```diff
@@ -0,0 +1,58 @@
...
+	public void getWindowHandleInNewWindow() {
+		pageElement = PageManager.setPage("HomePage");
+		baseUrl = Configuration.url();
+		PageManager.openPage(baseUrl);
+		pageElement.getElement("open_new_window").click();
+		PageManager.switchToNewWindow();
+		pageElement = PageManager.setPage("OpenNewWindowPage");
+		String expectedVal = "A New Popup Window";
+		String actualVal = pageElement.getElement("header_text1").getText();
+		Assert.assertTrue(actualVal.equals(expectedVal));
```
dougnoel, line 43: You will get a better error message when this fails, also since you are comparing two values, you should give them as separate arguments and use AssertEquals instead of AssertTrue. You can then lass in the expected argument, then the actual value, and if the value doesn't match the expected it will be clear in the unit test failure.
```suggestion
		Assert.assertEquals("Expected a new popup window to appear.", expectedVal, actualVal);
```

src/test/java/com/dougnoel/sentinel/pages/PageManagerTests.java
```diff
@@ -0,0 +1,58 @@
...
+	public void getWindowHandleInNewTab() {
+		pageElement = PageManager.setPage("HomePage");
+		baseUrl = Configuration.url();
+		PageManager.openPage(baseUrl);
+		pageElement.getElement("open_new_tab").click();
+		PageManager.switchToNewWindow();
+		pageElement = PageManager.setPage("OpenNewTabPage");
+		String expectedVal = "Window Opened in a New Tab";
+		String actualVal = pageElement.getElement("header_text2").getText();
+		Assert.assertTrue(actualVal.equals(expectedVal));
```
dougnoel, line 56: You will get a better error message when this fails, also since you are comparing two values, you should give them as separate arguments and use AssertEquals instead of AssertTrue. You can then lass in the expected argument, then the actual value, and if the value doesn't match the expected it will be clear in the unit test failure.
```suggestion
		Assert.assertEquals("Expected new tab to be opened.", expectedVal, actualVal);
```

dougnoel (commented):
Looks better. Few small changes.

dougnoel (changes requested):
Meant to mark the last one as request changes.

src/test/java/com/dougnoel/sentinel/pages/SampleTestPage.yml
dougnoel: Ok, then add this to a package called com.encodedna in src/test/java

src/test/java/com/dougnoel/sentinel/pages/PageManagerTests.java
```diff
@@ -0,0 +1,58 @@
...
+		driver.quit();
+	}
+	
+	@Test(expected = com.dougnoel.sentinel.exceptions.NoSuchWindowException.class)
+	public void getWindowHandleInSameWindow() {
+		PageManager.switchToNewWindow();
+	}
+	
+	@Test
+	public void getWindowHandleInNewWindow() {
```
dougnoel, line 34: This should be done through a Cucumber Step.

src/test/java/com/dougnoel/sentinel/pages/PageManagerTests.java
```diff
@@ -0,0 +1,58 @@
...
+		pageElement.getElement("open_new_window").click();
+		PageManager.switchToNewWindow();
+		pageElement = PageManager.setPage("OpenNewWindowPage");
+		String expectedVal = "A New Popup Window";
+		String actualVal = pageElement.getElement("header_text1").getText();
+		Assert.assertTrue(actualVal.equals(expectedVal));
+	}
+	
+	@Test
+	public void getWindowHandleInNewTab() {
```
dougnoel, line 47: This should be done through a Cucumber Step.

src/test/java/com/dougnoel/sentinel/pages/PageManagerTests.java
```diff
@@ -0,0 +1,58 @@
...
+	}
+
+	@AfterClass
+	public static void tearDownAfterClass() throws Exception {
+		driver.quit();
+	}
+	
+	@Test(expected = com.dougnoel.sentinel.exceptions.NoSuchWindowException.class)
+	public void getWindowHandleInSameWindow() {
+		PageManager.switchToNewWindow();
```
dougnoel, line 30: There is no assertion here. This should be done through a Cucumber Step.

dougnoel (changes requested):

dougnoel (approved):

PR #231, opened 2021-06-02
Added script to support video recording add to extent report

mdsapon:


pom.xml
```diff
@@ -101,6 +101,8 @@
     <sonar.projectKey>dougnoel_sentinel</sonar.projectKey>
     <sonar.organization>dougnoel</sonar.organization>
     <sonar.host.url>https://sonarcloud.io</sonar.host.url>
+    
+    <screen-recorder.version>0.7.7.0</screen-recorder.version>
```
dougnoel, line 105: Move this up to line 100 with the rest of the version numbers.

src/main/java/com/dougnoel/sentinel/steps/BaseSteps.java
```diff
@@ -172,4 +175,10 @@ public static void pressBrowserButton(String option) {
         }
     }
     
-}
\ No newline at end of file
+    @Then("^video recording (.*?)( was not)?(?: was)? attached to the report?$")
```
dougnoel, line 178: elementName will never contain "not"
```suggestion
    @Then("^video recording was( not)? attached to the report?$")
```

src/main/java/com/dougnoel/sentinel/steps/BaseSteps.java
```diff
@@ -172,4 +175,10 @@ public static void pressBrowserButton(String option) {
...
+    public static void addVideoToReport(String elementName, String assertion) {
```
dougnoel, line 179:
```suggestion
    public static void addVideoToReport(String assertion) {
```

src/main/java/com/dougnoel/sentinel/steps/BaseSteps.java
```diff
@@ -172,4 +175,10 @@ public static void pressBrowserButton(String option) {
...
+    	if(Configuration.toString("addVideoToReport")!=null && !elementName.contains("not"))
```
dougnoel, line 180: If the step exists, you shouldn't have to turn it on in the commandline. Ideally this would happen whether there was a cucumber step or not by adding it to the after step for each method.
```suggestion
    	if(Configuration.toString("addVideoToReport")!=null && !assertion.contains("not"))
```

src/main/java/com/dougnoel/sentinel/steps/BaseSteps.java
```diff
@@ -172,4 +175,10 @@ public static void pressBrowserButton(String option) {
...
+     	   ExtentCucumberAdapter.addTestStepLog("<a href='"+System.getProperty("user.dir")+"\\reports\\extent-cucumber.avi'>Click to see video recording</a>");
+
```
dougnoel, line 182: delete extra line

src/test/java/features/47 Swag Labs Login.feature
```diff
@@ -13,10 +13,13 @@ Feature: 47 Swag Labs Login
...
   @47B		
   Scenario: 47B Failed Login with Locked Out User
 	Given I am on the Sauce Demo Login Page
   	When I fill the account information for account LockedOutUser into the Username field and the Password field
   	  And I click the Login Button
-  	Then I verify the Error Message contains the text "Sorry, this user has been locked out."
\ No newline at end of file
+  	Then I verify the Error Message contains the text "Sorry, this user has been locked out."
+  	  And video recording was not attached to the report
+
```
dougnoel, line 25: delete extra line

src/test/java/features/47 Swag Labs Login.feature
```diff
@@ -13,10 +13,13 @@ Feature: 47 Swag Labs Login
   	  And I click the Login Button
   	Then I am redirected to the Sauce Demo Main Page
   	  And I verify the App Logo exists
+  	  And video recording was attached to the report
```
dougnoel, line 16: See if you can add a @record tag at the top of this scenario and then read it in the @After step in the TestRunner instead of doing a Cucumber step here.

src/test/java/tests/SentinelTests.java
```diff
@@ -20,14 +19,21 @@
...
+  //, tags = { "@47" }
 )
 
 public class SentinelTests {
     private static final Logger log = LogManager.getLogger(SentinelTests.class); // Create a logger.
 
+    @BeforeClass
+    public static void setUpBeforeClass() {
+        WebDriverFactory.instantiateWebDriver();
+        if(Configuration.toString("addVideoToReport")!=null)
```
dougnoel, line 31: Put this in a @Before step that will run before each Scenario, and then see if you can read a @record tag for the Cucumber step to turn this on. Then this would be on if either the `addVideoToReport` option is on OR a scenario is tagged AND the record is not explicitly set to off.

If @record tag exists and `addVideoToReport` is not set (null) = record
If @record tag doesn't exist and `addVideoToReport` is set true = record
If @record tag exists and `addVideoToReport` is set true = record
If @record tag exists and `addVideoToReport` is set off/false = DON'T record
Else = DON'T record

src/test/java/tests/SentinelTests.java
```diff
@@ -36,5 +42,7 @@ public static void tearDownAfterClass() throws SentinelException {
         if (System.getProperty("leaveBrowserOpen", "false") == "false") {
         	PageManager.quit();
         }
+        if(Configuration.toString("addVideoToReport")!=null)
```
dougnoel, line 45: Move this to an @After method as I mentioned in previous comments.

dougnoel (changes requested):

PR #233, opened 2021-06-04
Completed # 3049 - When the "elements:" line is missing from the yaml file we should throw a ConfigurationParseException with a message

paglajewel:


src/main/java/com/dougnoel/sentinel/pages/PageData.java
```diff
@@ -91,8 +92,14 @@ public static PageData loadYaml(File fileName) throws IOException{
      * @return Map&lt;String, String&gt; the locators for an element
      */
     public Map<String,String> getElement(String elementName) {
-    	if (elements.containsKey(elementName)) {
-    		return elements.get(elementName);
+    	String pageName = PageManager.getPage().getName();
```
dougnoel, line 95: Move this  into line 101 and don't declare a variable, just put the call to PageManager.getPage().getName() inline.

src/test/java/com/dougnoel/sentinel/pages/PageDataTests.java
```diff
@@ -0,0 +1,47 @@
...
+	@AfterClass
+	public static void tearDownAfterClass() throws Exception {
+		driver.close();
+	}
+	
+	@Test
+	public void validateElementsExistsInYml() {
+		PageManager.setPage("PageObjStructure1");
+		String pageName = PageManager.getPage().getName();
+		Configuration.getElement("user_div", pageName);
```
dougnoel, line 31: What's the assertion here?

dougnoel (changes requested):

src/main/java/com/dougnoel/sentinel/pages/PageData.java
paglajewel: done

src/test/java/com/dougnoel/sentinel/pages/PageDataTests.java
paglajewel: done

dougnoel (changes requested):
Looking at code coverage on line 95 of PageData.java, we only cover 3 of 4 possible scenarios. Is it possible to cover the fourth?

https://sonarcloud.io/component_measures?id=dougnoel_sentinel&metric=new_coverage&pullRequest=233&selected=dougnoel_sentinel%3Asrc%2Fmain%2Fjava%2Fcom%2Fdougnoel%2Fsentinel%2Fpages%2FPageData.java&view=list

dougnoel (changes requested):
Let's see if we can improve code coverage. Otherwise, looks good.

dougnoel:
100% code coverage on new code attained. The remaining code smell requires an upgrade to JUnit 5 or move to TestNG, both of which are out of scope for this ticket.

PR #235, opened 2021-06-11
197 add an on hover test step1

dougnoel:


src/main/java/com/dougnoel/sentinel/elements/PageElement.java
```diff
@@ -199,6 +199,16 @@ public PageElement javaScriptSendKeys(String text) {
 
 		return this;
 	}
+	
+	/**
+	 * Moves the mouse to the middle of the element.	
+	 * @return PageElement (for chaining)
+	 */
+	
+	public PageElement mouseOver() {
```
dougnoel, line 208: Change this to a method that gives us something actionable. Call it getTooltip. Then do the mouseover and return the tooltip instead of the PageElement object. We could do these as separate steps, but it currently gets us nothing, so make this a terminal method.

src/main/java/com/dougnoel/sentinel/steps/BaseSteps.java
```diff
@@ -46,7 +52,23 @@ public static void before(Scenario scenario) {
...
+     * <p>
+     * <b>Gherkin Examples:</b>
+     * <ul>
+     * <li>When I mouse over element</li>
+     * </ul> 
+     * @param elementName String the target element to move to
+     */
+    
+     @When("^I mouse over (.*?)$") 
+    public void mouseoveronTarget(String elementName) {
```
dougnoel, line 68: Remove this method from this file and instead update the TextVerificationSteps and add the ability to check the tooltip text of an item as part of the text equals/contains step.

src/test/java/features/MouseHover.feature
```diff
@@ -0,0 +1,18 @@
...
+  And click on the Hovers
+  And Mouse Hover on User1, Uesr2,User3
+
+  @197A
+  Scenario: Successfully was able to Mouse Over on User1, User2, User3
+    Given I am on the Internet Herokuapp Page
+     When I click the hovers link   
+     Then I am redirected to the Hovers Page
+     When I mouse over User Div
+
```
dougnoel, line 15: Delete blank lines at the end of files.

src/test/java/features/MouseHover.feature
```diff
@@ -0,0 +1,18 @@
+﻿#language: en
+#Author: Shobana Namburaj
+@197
+Feature: Heroku App Mouse Hover
+  Login to the HerokuApp 
+  And click on the Hovers
+  And Mouse Hover on User1, Uesr2,User3
+
+  @197A
+  Scenario: Successfully was able to Mouse Over on User1, User2, User3
```
dougnoel, line 10: Update this method to validate the text of something we can mouse over.

src/test/java/pages/HoversPage.yml
```diff
@@ -0,0 +1,6 @@
+urls:
+   default: http://the-internet.herokuapp.com/hovers
+   demo: http://localhost:3001
```
dougnoel, line 3: Get rid of the demo entry.

src/test/java/pages/InternetHerokuappPage.yml
```diff
@@ -0,0 +1,10 @@
+urls:
+   default: http://the-internet.herokuapp.com/
+   demo: http://localhost:3001
```
dougnoel, line 3: Get rid of the demo entry.

src/test/java/pages/InternetHerokuappPage.yml
```diff
@@ -0,0 +1,10 @@
+urls:
+   default: http://the-internet.herokuapp.com/
```
dougnoel, line 2: I believe this file already exists. You should just be adding a new element to it.

dougnoel (commented):
Requesting changes. Also merge in master to fix the unit tests passing.

src/main/java/com/dougnoel/sentinel/elements/PageElement.java
roshna0104: Updated the existing method mouseOver() with the target element and added a new method getToolTipText() to return the text

src/main/java/com/dougnoel/sentinel/steps/BaseSteps.java
roshna0104: I updated the method to get the tool tip text from the page element

src/test/java/features/MouseHover.feature
roshna0104: Done

src/test/java/features/MouseHover.feature
roshna0104: Done

src/test/java/pages/HoversPage.yml
roshna0104: Done

src/test/java/pages/InternetHerokuappPage.yml
roshna0104: Done

src/test/java/pages/InternetHerokuappPage.yml
roshna0104: I don't see any duplicate in my local.

src/main/java/com/dougnoel/sentinel/elements/PageElement.java
```diff
@@ -6,6 +6,7 @@
 import java.time.Duration;
```
dougnoel, line 6: You ran this file through a coding standards tool, and while I appreciate the thought there are some things in here I agree with and some I don't. Please revert the file and only check in your changes. Running code through a tool like this requires discussion. This file is a core file and I need to go through it with a fine-tooth comb. That is difficult to do when there are so many spacing changes distracting from your code. If you want to do a cleanup like this, we should first discuss the rules, and second these changes should be made in a PR without any functional changes.

src/main/java/com/dougnoel/sentinel/steps/BaseSteps.java
```diff
@@ -1,8 +1,10 @@
 package com.dougnoel.sentinel.steps;
 
 import static com.dougnoel.sentinel.elements.ElementFunctions.getElement;
+
```
dougnoel, line 4: I'm not sure what happened to this file, but there are way too many changes in it unrelated to what you were working on. Please revert it and only make the changes necessary for your code. This is too much to review.

src/main/java/com/dougnoel/sentinel/steps/TextVerificationSteps.java
```diff
@@ -16,172 +16,197 @@
 import io.cucumber.java.en.Then;
```
dougnoel, line 16: Please revert this file and only add your changes.

src/test/java/pages/HoversPage.yml
```diff
@@ -0,0 +1,7 @@
+urls:
+   default: http://the-internet.herokuapp.com/hovers  
+elements: 
+  user_div:
+   xpath: "(//img[@alt='User Avatar'])[1]"  
+  tooltip:
```
dougnoel, line 6: The tooltip xpath should no longer be in here.

src/test/java/pages/InternetHerokuappPage.yml
dougnoel: src/test/java/com.herokuapp/InternetPage.yml

This file was not created to the correct specs. It should be in a package matching the production URL.

dougnoel (commented):
Additional changes needed. Please revert changes that were not part of the coding you did. I appreciate the interest in cleaning the file up to meet a standard, but we have not agreed on a standard and those changes should be made in a different pull request.

src/test/java/pages/InternetHerokuappPage.yml
dougnoel: This still needs to be deleted and the correct page used.

src/test/java/pages/HoversPage.yml
dougnoel: This needs to be deleted. You cannot have a declared xpath to the tooltip or this is going to take too much work to use.

src/test/java/features/MouseHover.feature
```diff
@@ -0,0 +1,16 @@
...
+#Author: Shobana Namburaj
+@197
+Feature: Heroku App Mouse Hover
+  Login to the HerokuApp 
+  And click on the Hovers
+  And Mouse Hover on User Profile and validate the tooltip text
+
+  @197A
+  Scenario: Successfully was able to Mouse Over and validate the text "name: user1"
+    Given I am on the Internet Herokuapp Page
```
dougnoel, line 11:
```suggestion
    Given I am on the Internet Page
```

src/main/java/com/dougnoel/sentinel/steps/TextVerificationSteps.java
```diff
@@ -81,17 +83,17 @@ public static void verifyElementTextContains(String elementName, String assertio
             String elementText = getElement(elementName).getText();
             var expectedResult = SentinelStringUtils.format(
                     "Expected the {} element to {}{} the text {}. The element contained the text: {}",
```
dougnoel, line 85: This is needed for the hasTooltipText you are passing below. Otherwise you will end up with the message being incorrect.
```suggestion
                    "Expected the {} element{} to {}{} the text {}. The element contained the text: {}",
```

src/main/java/com/dougnoel/sentinel/steps/TextVerificationSteps.java
```diff
@@ -68,10 +68,12 @@ public static void verifyElementTextIsEmpty(String elementName, String assertion
...
      * @param text String The text to verify exists in the element.
      */
-    @Then("^I verify the (.*?)( does not)? (has|have|contains?) the text \"([^\"]*)\"$")
-    public static void verifyElementTextContains(String elementName, String assertion, String matchType, String text) {
+    @Then("^I verify the (.*?)( tooltip)?( does not)? (has|have|contains?) the text \"([^\"]*)\"$")
+    public static void verifyElementTextContains(String elementName, String assertion, String tooltip, String matchType, String text) {
         boolean negate = !StringUtils.isEmpty(assertion);
+        boolean hasTooltip = !StringUtils.isEmpty(tooltip);
         String negateText = negate ? "not " : "";
+        String hasTooltipText = hasTooltip ? "not " : "";
```
dougnoel, line 76: Just putting the word tooltip or nothing.
```suggestion
        String hasTooltipText = hasTooltip ? "tooltip " : "";
```

src/main/java/com/dougnoel/sentinel/steps/TextVerificationSteps.java
```diff
@@ -81,17 +83,17 @@ public static void verifyElementTextContains(String elementName, String assertio
             String elementText = getElement(elementName).getText();
             var expectedResult = SentinelStringUtils.format(
                     "Expected the {} element to {}{} the text {}. The element contained the text: {}",
-                    elementName, negateText, partialMatchText, text, elementText
+                    elementName, negateText, hasTooltipText, partialMatchText, text, elementText
                             .replace("\n", " "));
             log.trace(expectedResult);
             if (partialMatch) {
-                if (negate) {
+                if (negate && hasTooltip) {
```
dougnoel, line 90: This only works now if you have a tooltip. This still needs to work if there is no tooltip.

src/main/java/com/dougnoel/sentinel/steps/TextVerificationSteps.java
```diff
@@ -81,17 +83,17 @@ public static void verifyElementTextContains(String elementName, String assertio
...
                     assertFalse(expectedResult, elementText.contains(text));
                 } else {
                     assertTrue(expectedResult, elementText.contains(text));
                 }
             } else {
-                if (negate) {
+                if (negate && hasTooltip) {
```
dougnoel, line 96: This only works now if you have a tooltip. This still needs to work if there is no tooltip.

src/test/java/tests/SentinelTests.java
```diff
@@ -21,7 +21,7 @@
 	, plugin = {"json:target/cucumber.json",
 			"com.aventstack.extentreports.cucumber.adapter.ExtentCucumberAdapter:"}
 	, strict = true
-//  , tags = { "@216" }
+  , tags = { "@197" }
```
dougnoel, line 24: This has to be commented out or it's the only unit test run.

src/test/java/tests/SentinelTests.java
```diff
@@ -11,8 +11,8 @@
 import com.dougnoel.sentinel.pages.PageManager;
 import com.dougnoel.sentinel.webdrivers.WebDriverFactory;
 
-import io.cucumber.junit.CucumberOptions;
```
dougnoel, line 14: If you don't change this file at all on check-in, it doesn't need to be added.

dougnoel (commented):
Looking much better. Still some changes needed.

dougnoel:
> 4 layers of condition made it little complex to read, check with Doug if we can simplify this complexity

I agree with Mizanur. We may want to separate out these conditions into helper methods.

Issue #292, opened 2022-02-22
Add ability to send special keys to element


sampacos:
In some scenarios, a user needs to be able to send special keys (such as backspace) without any other validations or interactions.

Issue #291, opened 2022-02-22
Table getElementInRowThatContains(int, By) searches table structure incorrectly for element


sampacos:
`getElementInRowThatContains(int, By)` chains an extra `.findElement` that is not necessary. This extra call is for any element matching the xpath "//td", which matches the first td element in the entire document. This invalidates the following `.findElement` call. Removing lines 315 and 326 will fix this issue.

An additional problem in Table.java: `getOrCreateRowElements()` has the potential to find rows in the header. This means the row indexing for multiple methods is off. Changing the selector to only look in the tbody of the current table element would solve this issue.

PR #293, opened 2022-02-22
291 table fixes and 292 add element method for sending special keys

sampacos:
#291 #292 
Fixes for a couple table methods to make fetching elements in rows more reliable.
Add method in Element to send a special character without any other interactions or verifications.


dougnoel (approved):
Reviewed in meeting.

Issue #305, opened 2022-03-16
Add verification of specific rows in tables


sampacos:
Current table verification steps allow for verification that "any" or "all" rows of a table contain specific text in a specific column. I need this ability for specific rows i.e. 1st, 2nd, last, etc.

sampacos:
closed with #306

PR #306, opened 2022-03-16
add table verification step for getting specific cell data

sampacos:
#305 
- add Table Verification step for getting specific cell data
- edit method in table to fetch column data faster using xpath. 
- add usage of step in table steps feature to cover new code.

### DIFF
--- a/src/main/java/com/dougnoel/sentinel/elements/dropdowns/SelectElement.java
+++ b/src/main/java/com/dougnoel/sentinel/elements/dropdowns/SelectElement.java
@@ -122,4 +122,13 @@ public class SelectElement extends Element {
 				.ignoring(StaleElementReferenceException.class)
 				.until(d -> selectElement.getOptions().stream().noneMatch(el -> el.getText().endsWith(text)));
     }
+    
+    /**
+     * Returns the number of options in this Select element.
+     * @return int the number of options in this Select element.
+     */
+    public int getNumberOfOptions() {
+    	Select selectElement = new Select(this.element());
+    	return selectElement.getOptions().size();
+    }
 }
--- a/src/main/java/com/dougnoel/sentinel/elements/dropdowns/SelectElement.java
+++ b/src/main/java/com/dougnoel/sentinel/elements/dropdowns/SelectElement.java
@@ -128,7 +128,6 @@ public class SelectElement extends Element {
      * @return int the number of options in this Select element.
      */
     public int getNumberOfOptions() {
-    	Select selectElement = new Select(this.element());
-    	return selectElement.getOptions().size();
+    	return new Select(this.element()).getOptions().size();
     }
 }
--- a/src/main/java/com/dougnoel/sentinel/elements/tables/Table.java
+++ b/src/main/java/com/dougnoel/sentinel/elements/tables/Table.java
@@ -398,9 +398,15 @@ public class Table extends Element {
 	 * @return ArrayList&lt;String&gt; a list of all cell data. Each entry in the list corresponds to a cell.
 	 */
 	public List<String> getAllCellDataForColumn(String columnHeader){
-		int indexOfHeader = getOrCreateHeaders().indexOf(columnHeader) + 1; // add 1 because the List.getIndex method is 0-indexed and XPath is 1-indexed
+		int arrayIndexOfHeader = getOrCreateHeaders().indexOf(columnHeader);
+		if(arrayIndexOfHeader == -1) {
+			String errorMessage = SentinelStringUtils.format("{} column does not exist.", columnHeader);
+			log.error(errorMessage);
+			throw new NoSuchElementException(errorMessage);
+		}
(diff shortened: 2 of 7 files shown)


src/main/java/com/dougnoel/sentinel/steps/TableVerificationSteps.java
```diff
@@ -186,5 +188,38 @@ public static void verifyColumnSort(String columnName, String tableName, String
...
+     * @param columnName String the name of the column to verify
+     * @param tableName String the name of the table containing the column
+     * @param assertion String if null is passed, looks for match(es), if any strong value is passed, looks for the value to not exist.
+     * @param textToMatch String the text to look for in the column
+     */
+    @Then("^I verify the cell in the (la|\\d+)(?:st|nd|rd|th) row and the (.*) column of the (.*?)( do(?:es)? not)? contains? the text (.*?)$")
+    public static void verifyCellInSpecifiedRow(String rowNum, String columnName, String tableName, String assertion, String textToMatch) {
+    	boolean negate = !StringUtils.isEmpty(assertion);
+    	
+    	var table = (Table)getElement(tableName);
```
dougnoel, line 210: The method for returning this wraps the error if it can't be cast, but I'm starting to wonder if we should instead get rid of that code and use the way you've done it here. If you feel strongly about the way you have this here we can discuss it.
```suggestion
    	var table = getElementAsTable(tableName);
```

src/test/java/features/Table Tests.feature
```diff
@@ -24,6 +25,9 @@ Feature: Table Tests
       And I verify the cells in the Distance column in the Example Table are sorted in ascending order
       And I verify the cells in the ID Number column in the Example Table are sorted in descending order
       And I view the 1st page of results from the Example Table
+      And I verify the cell in the 1st row and the First Name column of the Example table contains the text Bob
```
dougnoel, line 28: Can we test the 2nd row here for more coverage of the regex?

src/test/java/features/Table Tests.feature
```diff
@@ -24,6 +25,9 @@ Feature: Table Tests
...
+      And I verify the cell in the 1st row and the First Name column of the Example table does not contain the text Charlotte
```
dougnoel, line 29: Can we test the 3rd row here for more coverage of the regex?

src/test/java/features/Table Tests.feature
```diff
@@ -10,6 +10,7 @@ Feature: Table Tests
     When I click the Name header
     Then I verify the cells in the Name column in the Example Table are sorted in ascending order
       And I see 100 rows in the Example Table
+      And I verify the cell in the 1st row and the Name column of the Example table contains the text Alexander Foley
```
dougnoel, line 13: These are good but I'd also like to make sure they work with the NGXDataTable. CAn you add the same tests to that feature file?

src/main/java/com/dougnoel/sentinel/elements/tables/Table.java
```diff
@@ -398,8 +398,10 @@ public void clickElementInRowThatContains(int ordinalRow, By elementLocator) {
 	 * @return ArrayList&lt;String&gt; a list of all cell data. Each entry in the list corresponds to a cell.
 	 */
 	public List<String> getAllCellDataForColumn(String columnHeader){
-		getOrCreateHeaders();
-		return getOrCreateColumns().get(columnHeader);
+		int indexOfHeader = getOrCreateHeaders().indexOf(columnHeader) + 1;
```
dougnoel, line 401: I don't love comments typically in code, but in this case I'd actually like you to add a comment about why the +1 is being added. This is one of those things that is very confusing coming into it.

dougnoel (changes requested):

sampacos:
Test failure is a result of inconsistent loading of the NGX Data Table Page. When the window is not maximized, the table on the page sometimes loads without the 3rd column.

src/main/java/com/dougnoel/sentinel/elements/dropdowns/SelectElement.java
```diff
@@ -122,4 +122,13 @@ public boolean doesNotHaveOption(String text) {
...
 				.until(d -> selectElement.getOptions().stream().noneMatch(el -> el.getText().endsWith(text)));
     }
+    
+    /**
+     * Returns the number of options in this Select element.
+     * @return int the number of options in this Select element.
+     */
+    public int getNumberOfOptions() {
+    	Select selectElement = new Select(this.element());
+    	return selectElement.getOptions().size();
```
dougnoel, line 132: I may be over-optimizing, but we don't actually need the variable declaration.
```suggestion
    	return new Select(this.element()).getOptions().size();
```

dougnoel (changes requested):
Three more things.

dougnoel (approved):

PR #322, opened 2022-04-06
Enhance table html

sampacos:
add html buttons


sampacos:
accidentally made branch off of feature branch instead of off of master

PR #343, opened 2022-04-25
336 windows screenshot steps

tyBouch:
Followed some additional input from intellij's code sanitizer to update a bit of the changed code to follow a sensical standard (according to its check-in verification). let me know how it all looks. Not sure if the comment on the 255 should be kept.


src/main/java/com/dougnoel/sentinel/elements/WindowsElement.java
```diff
@@ -182,38 +182,22 @@ public Element click() {
...
 	 */
-	public Color getColor(int xOffset, int yOffset) {
-		Robot robot;
-		try {
-			robot = new Robot();
-		} catch (AWTException awte) {
-			throw new com.dougnoel.sentinel.exceptions.IOException(awte);
-		}
-
-		// Grab the pixel coordinates of the rectangle of the element, relative to the top-left corner of the screen. Positive X = right. Positive Y = down.
```
pturchinetz, line 193: I'm not saying that this needs to change, but I noticed in my click position work that the 'official' W3 standard is no longer to use the top left, and instead the center of the element. You should see what people suggest as best practice.

Note #4: https://www.w3.org/TR/webdriver/#wheel-actions

src/main/java/com/dougnoel/sentinel/elements/WindowsElement.java
```diff
@@ -182,38 +182,22 @@ public Element click() {
...
 		try {
-			ImageIO.write(screenshot, "png", outputfile);
+			FileManager.saveImage(null, "tempGetColorImage.png", screenshot);
 		} catch (IOException ioe) {
 			throw new com.dougnoel.sentinel.exceptions.IOException(ioe);
 		}
 
 		int argb = screenshot.getRGB(xOffset, yOffset);
-		double a = (argb >> 24) & 0xFF;
+		double a = ((argb >> 24) & 0xFF)/255.0; //The selenium color object will take any double for alpha, but requires 0.0-1.0 for correct operation. This converts the alpha from 0-255 to the correct range.
```
pturchinetz, line 197: Awesome catch here

src/main/java/com/dougnoel/sentinel/elements/WindowsElement.java
```diff
@@ -244,9 +228,62 @@ public Color getColor() {
...
+	 *
+	 * @return File a screenshot of the current element
+	 */
+	@Override
+	public File getScreenshot() {
+		BufferedImage temporaryBuffer = screenshotElement();
+		try{
+			return FileManager.saveImage(null, "tempScreenshotImage.png", temporaryBuffer);
+		}
+		catch(IOException e){
```
pturchinetz, line 258: Spacing inconsistent on this try{}catch(){

src/main/java/com/dougnoel/sentinel/elements/WindowsElement.java
```diff
@@ -244,9 +228,62 @@ public Color getColor() {
...
+	 */
+	private BufferedImage screenshotElement() {
+		Robot robot;
+		try {
+			robot = new Robot();
+		} catch (AWTException awte) {
+			throw new com.dougnoel.sentinel.exceptions.IOException(awte);
+		}
+
+		// Grab the pixel coordinates of the rectangle of the element, relative to the top-left corner of the screen. Positive X = right. Positive Y = down.
```
pturchinetz, line 276: Same comment with the offset guide changing to have the origin be the center

src/main/java/com/dougnoel/sentinel/steps/ImageSteps.java
```diff
@@ -27,22 +27,23 @@ public static void before(Scenario scenario) {
 	/**
 	 * Takes a screenshot of the given element and stores it for later comparison
 	 * @param elementName String name of the element to screenshot
-	 * @throws IOException
+	 * @throws IOException If the expected image to be created fails to save
 	 */
-    @When("^I take a screenshot of (?:the|a) (page|.*)$")
```
pturchinetz, line 32: I feel like the term page here is ambiguous with page object, but I can't think of a better term.

src/main/java/com/dougnoel/sentinel/steps/ImageSteps.java
```diff
@@ -27,22 +27,23 @@ public static void before(Scenario scenario) {
...
+    @When("^I take a screenshot of (?:the|a) (.*?)$")
     public static void storeScreenshotOfElement(String elementName) throws IOException {
-    	String imageFileName = scenario.getName()+ "_" + elementName + "_" + PageManager.getPage().getName() + ".png";
+		String outputFolder = "imageComparison/" + scenario.getName();
+    	String imageFileName = PageManager.getPage().getName() + "_" + elementName + "_EXPECTED" + ".png";
```
pturchinetz, line 35: Should there be an element of randomness (timestamp) in the name, or is it smart enough to overwrite on a rerun? Do we want to keep the old screenshots for records' sake?

src/main/java/com/dougnoel/sentinel/steps/ImageVerificationSteps.java
```diff
@@ -54,48 +56,64 @@ public static void verifyImageNotMatch(String elementName, String assertion) thr
...
 	 * 
 	 * @return ImageComparisonResult the image comparison result.
 	 */
     private static ImageComparisonResult compareImages(String elementName) throws IOException {
-    	String imageFileName = scenario.getName()+ "_" + elementName + "_" + PageManager.getPage().getName() + ".png";
-    	String resultImageFileName = scenario.getName()+ "_" + elementName + "_" + PageManager.getPage().getName() + ".png";
+		String outputFolder = "ImageComparison/" + scenario.getName();
+    	String actualFileName = PageManager.getPage().getName() + "_" + elementName + "_ACTUAL" + ".png";
+		String expectedFileName = PageManager.getPage().getName() + "_" + elementName + "_EXPECTED" + ".png";
+    	String resultImageFileName = PageManager.getPage().getName() + "_" + elementName + "_RESULT" + ".png";
```
pturchinetz, line 84: Same note on randomness in file names

src/main/java/com/dougnoel/sentinel/steps/ImageVerificationSteps.java
```diff
@@ -54,48 +56,64 @@ public static void verifyImageNotMatch(String elementName, String assertion) thr
...
+			}
     	}
     	else {
-    		TakesScreenshot pageScreenshotTool =((TakesScreenshot) WebDriverFactory.getWebDriver());
+    		TakesScreenshot pageScreenshotTool =((TakesScreenshot) Driver.getWebDriver());
     		screenshotFile = pageScreenshotTool.getScreenshotAs(OutputType.FILE);
-    		Element body = new Element("body", Map.of("xpath", "//body"));
-    		backgroundColor = body.getBackgroundColor();
+
+			if(PageManager.getCurrentPageObjectType() != PageObjectType.EXECUTABLE) {
```
pturchinetz, line 102: We should probably switch on PageManager.getCurrentPageObjectType() instead

src/main/java/com/dougnoel/sentinel/system/FileManager.java
```diff
@@ -155,6 +167,30 @@ public static BufferedImage readImage(String subDirectory, String fileName) {
...
+	 * @param fileToCheck File the file object of the file we wish to verify exists
+	 * @throws IOException in the case the file does not begin existing within the default timeout
+	 */
+	private static void verifyFileCreation(File fileToCheck) throws IOException {
+		var errorMessage = SentinelStringUtils.format("Failed to create the {} image file", fileToCheck.getAbsolutePath());
+
+		long searchTime = Time.out().getSeconds() * 1000;
+		long startTime = System.currentTimeMillis(); // fetch starting time
+		while ((System.currentTimeMillis() - startTime) < searchTime) {
+			if (Files.exists(fileToCheck.toPath()))
```
pturchinetz, line 190: Was this taking a long time to search? It feels odd to need a timeout on something like this.

src/test/java/features/239 Initial WinAppDriver Implementation.feature
```diff
@@ -45,4 +45,37 @@ Feature: 239 Implement WinAppDriver to automate windows
...
+  	Then I verify the Header contains the text "Window Opened in a New Tab"
+
+  @239E
+  Scenario: 239E Verify notepad window has changed after entering text using image comparison
+	Given I switch to the Notepad App
+	  And I click the Text Editor field
+	  And I clear the Text Editor field
+	  And I verify the Text Editor field is empty
+	  And I take a screenshot of the window
+    When I enter test in the Text Editor field
```
pturchinetz, line 57: tabbing

src/test/java/features/239 Initial WinAppDriver Implementation.feature
```diff
@@ -45,4 +45,37 @@ Feature: 239 Implement WinAppDriver to automate windows
...
+  Scenario: 239G Verify notepad text editor field element has changed after entering text using image comparison
+	Given I switch to the Notepad App
+	  And I click the Text Editor field
+	  And I take a screenshot of the Text Editor field
+	When I enter test in the Text Editor field
+	Then I verify the Text Editor field does not match the original image
+
+  @239H
+  Scenario: 239H Verify notepad text editor field element has not changed when no operations have occurred using image comparison
+    Given I switch to the Notepad App
```
pturchinetz, line 79: tabbing

pturchinetz (changes requested):

src/main/java/com/dougnoel/sentinel/elements/WindowsElement.java
tyBouch: Need to have a discussion on this as I believe we initially avoided middle since text color could be captured if we went dead center.

src/main/java/com/dougnoel/sentinel/elements/WindowsElement.java
tyBouch: Debugging that was so...that was still funny when I realized it.

src/main/java/com/dougnoel/sentinel/steps/ImageSteps.java
tyBouch: Problem is a user using windows writing tests. We don't necessarily anticipate the person writing the steps to be writing the tests. So it just made sense to allow page or window.

I don't really know many people who would call explorer.exe a page even if they were using a page object.

src/main/java/com/dougnoel/sentinel/elements/WindowsElement.java
tyBouch: Kinda confused what you mean.

src/main/java/com/dougnoel/sentinel/steps/ImageSteps.java
tyBouch: For right now it's smart enough to perform overwrites. Unless I'm horribly mistaken.

Archiving for later is not currently implemented, but we can look into it later.

src/main/java/com/dougnoel/sentinel/system/FileManager.java
tyBouch: It was not. The Time.out().getSeconds(), remember, just gets the number of seconds before aborting, not an actual pause. This'll operate as fast as it needs to.

It may also no longer be necessary, but I did see a failure due to file creation (It may have been a mad location at the time, though). If you want we can pair up and go over it to see if we can trim it out.

src/test/java/features/239 Initial WinAppDriver Implementation.feature
tyBouch: Yeah I really gotta fix that. Lets meet up and fix up intellij's spacing problems in feature files.

tyBouch:
Please do not merge as additional issues have been raised outside this review.

src/main/java/com/dougnoel/sentinel/steps/ImageVerificationSteps.java
```diff
@@ -54,48 +59,78 @@ public static void verifyImageNotMatch(String elementName, String assertion) thr
...
+				case WEBPAGE:
+					backgroundColor = getElement(elementName).getBackgroundColor();
+					break;
+				case UNKNOWN:
+				default:
+					log.warn("Page object type is of the unhandled {} type. Black will be used as background color fallback.", pageType);
+					backgroundColor = Color.BLACK;
+					break;
+			}
+		}
```
dougnoel, line 127: This code breaks encapsulation. A test step shouldn't need to know the implementation details of the objects it is comparing. The objects need a comparison feature we can call. If they don't have that, we should wrap them in our own object(s) and create that comparison. In that way, if we create a thrid screenshot implementation (say for a mobile device) this code doesn't need to change. Also, in theory, when I take a screenshot of a window and a web page rendering the same data I should be able to prove they match. While currently the implementation of the taking of the screenshot is different, the output should be the same.

src/main/java/com/dougnoel/sentinel/steps/ImageVerificationSteps.java
```diff
@@ -54,48 +59,78 @@ public static void verifyImageNotMatch(String elementName, String assertion) thr
...
+	 * will be set to parent background color for web, upper-left-most color for windows applications.
+	 * </i>
+	 *
 	 * @param elementName String the name of the element to capture and compare 
 	 * or any casing of "page" alone to compare the entire page.
 	 * @throws IOException if file creation does not work
 	 * 
 	 * @return ImageComparisonResult the image comparison result.
 	 */
     private static ImageComparisonResult compareImages(String elementName) throws IOException {
```
dougnoel, line 83: See additional comments below, but this method breaks encapsulation. It should exist inside of the screenshots themselves. If there are special cases to handle them you should wrap the screenshots and handle that.

src/main/java/com/dougnoel/sentinel/steps/ImageVerificationSteps.java
```diff
@@ -104,23 +139,22 @@ private static ImageComparisonResult compareImages(String elementName) throws IO
...
 
 		if((expectedWidth != actualWidth) || (expectedHeight != actualHeight)) {
+			switch(pageType){
+				case WEBPAGE:
+					break;
+				case EXECUTABLE:
+				case UNKNOWN:
+				default:
+					log.warn("Page object has been detected as {} with different sized comparison snapshots. This may result in an unreliable comparison.", pageType);
+					break;
```
dougnoel, line 149: It is unclear to me why the comparison wouldn't fail at this point for all types unless this is a "contains" instead of a "has".

src/test/java/features/239 Initial WinAppDriver Implementation.feature
```diff
@@ -3,46 +3,79 @@
...
+      And I click the Text Editor field
+      And I take a screenshot of the Text Editor field
+    When I enter test in the Text Editor field
+    Then I verify the Text Editor field does not match the original image
+
+  @239H
+  Scenario: 239H Verify notepad text editor field element has not changed when no operations have occurred using image comparison
+    Given I switch to the Notepad App
+      And I take a screenshot of the Text Editor Field
+    Then I verify the Text editor Field matches the original image
```
dougnoel, line 81: This is a lot of testing time for a couple validations. While I do like atomic tests, this level of granularity is not necessary in this case. We are specifically testing Our WinAppDriver tests. Combining these tests will create a real world example, and reduce the total test time significantly cause we don't have to keep re-opening Notepad. It will also test all your inherent assumptions about naming convention and make sure you don''t have any hidden name collision issues.

dougnoel (changes requested):

src/main/java/com/dougnoel/sentinel/elements/WindowsElement.java
```diff
@@ -244,9 +228,61 @@ public Color getColor() {
 	@Override
 	public boolean attributeEquals(String attribute, String value) {
 		if (attribute.equalsIgnoreCase("color")) {
-			var hexColor = getColor().asHex();
+			var hexColor = getColorAtOffset().asHex();
```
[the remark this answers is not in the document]
tyBouch: It's actually used in another area or two now, and by returning it as a color instead of hex we're able to use the selenium color converter to convert it between selenium and java color objects.

A change I want to make is to create a java color object, actually, rather than a selenium color, though I think that also makes conversion to hex harder. So it's something to discuss once we implement the image comparison in a different area.

src/main/java/com/dougnoel/sentinel/elements/WindowsElement.java
```diff
@@ -244,9 +228,61 @@ public Color getColor() {
...
 			return hexColor.equalsIgnoreCase(value);
 		}
 		return super.attributeEquals(attribute, value);
 	}
-}
+
+	/**
+	 * Windows does not support color information, as such this is an unsupported operation for windows elements.
+	 *
+	 * @return NotImplementedException
```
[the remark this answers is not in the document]
tyBouch: Handled.

src/main/java/com/dougnoel/sentinel/elements/WindowsElement.java
```diff
@@ -244,9 +228,61 @@ public Color getColor() {
...
+	 */
+	@Override
+	public java.awt.Color getBackgroundColor() throws NotImplementedException {
```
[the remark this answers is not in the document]
tyBouch: I'll see about wrapping the exception so we can use the sentinel throw.

src/main/java/com/dougnoel/sentinel/elements/WindowsElement.java
```diff
@@ -244,9 +228,61 @@ public Color getColor() {
...
+	 * Returns a BufferedImage taken of the current element
+	 *
+	 * @return BufferedImage A robot screenshot cropped to the current element
+	 */
+	private BufferedImage screenshotElement() {
+		Robot robot;
+		try {
+			robot = new Robot();
+		} catch (AWTException awte) {
+			throw new com.dougnoel.sentinel.exceptions.IOException(awte);
```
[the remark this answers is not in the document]
tyBouch: Handled!

src/main/java/com/dougnoel/sentinel/system/FileManager.java
```diff
@@ -155,6 +165,11 @@ public static BufferedImage readImage(String subDirectory, String fileName) {
 	* @return String the constructed path as a String
 	*/
 	private static String getImagePath(String subDirectory, String fileName) {
-		return Configuration.toString("imageDirectory", "logs/images").replace("/", File.separator) + File.separator + subDirectory + File.separator + fileName;
+		String outputSubDir = "";
+
+		if(subDirectory != null){
+			outputSubDir = subDirectory;
+		}
+		return Configuration.toString("imageDirectory", "logs/images").replace("/", File.separator) + File.separator + outputSubDir + File.separator + fileName;
```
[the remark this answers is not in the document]
tyBouch: Handled

src/test/java/features/239 Initial WinAppDriver Implementation.feature
tyBouch: We'll need to change how we handle screenshots. As of right now if these were combined we'd overwrite the previous screenshot's results each time a comparison occured.

src/main/java/com/dougnoel/sentinel/steps/WindowAndTabSteps.java
```diff
@@ -41,7 +41,7 @@ public static void closeCurrentWindow() {
     @Then("^I verify a new (?:tab|window) opens(?: to)? the (.*?)$")
     public static void openNewWindow(String pageName) throws InterruptedException {
         PageManager.setPage(pageName);
-        Driver.goToMostRecentWindow();
+        Driver.goToNewWindow();
```
dougnoel, line 44:
```suggestion
        Driver.waitForNewWindow();
```

dougnoel (dismissed):

Issue #327, opened 2022-04-12
Add step definitions for verifying file downloads


sampacos:
As a user I want to be able to write gherkin that allows me to verify a file of a specific filetype (extension) is downloaded

PR #345, opened 2022-05-05
327 download steps

### LINKED ISSUE #344: Duplicate hover step
`BaseSteps.hover()` and `BaseSteps.hoverOverElement()` do the same thing. One needs to get removed.

Also, there are step methods that are not declared as "static". This is a bad practice, because those steps cannot then be called by other steps.

sampacos:
Address #327 
Fix #344 
Add element function for getting attribute


src/main/java/com/dougnoel/sentinel/steps/DownloadVerificationSteps.java
```diff
@@ -0,0 +1,36 @@
...
+	public static void verifyFileWithExtensionDownloaded(String fileExtension) throws InterruptedException, IOException {
+		DownloadManager.setFileExtension(fileExtension);
+		var fileName = DownloadManager.monitorDownload();
+		String expectedResult = SentinelStringUtils.format("Expected a new file with the {} extension to be downloaded. "
+				+ "Perhaps the download did not complete in time. Check your timeout.",
+                fileExtension);
+		
+		assertTrue(expectedResult, !StringUtils.isEmpty(fileName));
+	}
+}
```
pturchinetz, line 36: I wrote a similar function to verify a file is downloaded by filename. I'll want to put my function on this page, we can coordinate that later. Are there other download functions we might need?

src/main/java/com/dougnoel/sentinel/system/DownloadManager.java
```diff
@@ -62,17 +67,32 @@ public static boolean isFileDownloaded(String fileName) {
...
 
+    /**
+     * Monitors the current set download directory and returns a filename once the
+     * file is downloaded.
+     * 
+     * @return String The name of the file that was downloaded.
+     * @throws InterruptedException if the file download is interrupted
+     * @throws IOException if the file cannot be created.
+     */
+    public static String monitorDownload() throws InterruptedException, IOException {
```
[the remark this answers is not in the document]
pturchinetz: Like, I would want to be able to monitor download for a download where I have the file name already, you know?

src/main/java/com/dougnoel/sentinel/system/DownloadManager.java
```diff
@@ -178,6 +198,16 @@ public static boolean verifyPDFContent(URL url, String expectedText, int pageSta
         }
         pdfStripper.setStartPage(pageStart);
         pdfStripper.setEndPage(pageEnd);
+
+        try {
+            pdDoc = PDDocument.load(file);
+        } catch (InvalidPasswordException e) {
+        	var errorMessage = SentinelStringUtils.format("PDF file {} was password protected.", url.toString());
```
pturchinetz, line 205: Could be overkill but can we pass the password to open it, if we know it?

sampacos: I actually don't know. I can check on that later

src/main/java/com/dougnoel/sentinel/steps/DownloadVerificationSteps.java
sampacos: I think we can add those as we think of them, but I don't have anything in mind right now.

src/main/java/com/dougnoel/sentinel/strings/SentinelStringUtils.java
```diff
@@ -36,4 +37,24 @@ public static String stripSurroundingQuotes(String text) {
     	return text;
     }
 
+	/**
+	 * Constructs the ordinal string of the given int. For example "1st", "2nd", "111th"
+	 * @param i int the integer to create an ordinal string for
+	 * @return String the ordinal string
+	 */
+	public static String ordinal(int i) {
```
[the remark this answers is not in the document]
sampacos: Yeah I added it to make logging for that sort of thing easier, but mostly in Salvo. I can add it where possible in sentinel.

src/main/java/com/dougnoel/sentinel/system/DownloadManager.java
sampacos: you actually can but I'm not going to implement it now

src/main/java/com/dougnoel/sentinel/system/DownloadManager.java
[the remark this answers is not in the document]
sampacos: I added a step in DownloadVerificationSteps that allows for verification of specific filenames getting downloaded

dougnoel (changes requested):

PR #352, opened 2022-05-18
Updating the navigation for the dropdown page.

plmlkrz:
Fixes for the Dropdown navigation page.


src/test/java/features/Dropdown Tests.feature
```diff
@@ -15,7 +15,12 @@ Feature: Dropdown Tests
 	
   @22
   Scenario: 22 Prime NG Dropdown
-    Given I am on the Prime NG Dropdown Page
+    Given I am on the Prime NG Home Page
+    When I click the Components Button
+    Then I am redirected to the Prime NG Setup Page
+      And I verify the Get Started Header exists
```
dougnoel, line 21: We do not need this here. The framework automatically waits for the page to load, and we aren't trying to test this. It's an extraneous test.

src/test/java/org/primefaces/PrimeNGSetupPage.yml
```diff
@@ -0,0 +1,11 @@
+urls:
+  default: "https://www.primefaces.org/primeng/setup"
+
```
dougnoel, line 3: This isn't necessary if we can't go directly to this page. If we can then we should be going directly here.

src/test/java/features/Dropdown Tests.feature
```diff
@@ -15,7 +15,12 @@ Feature: Dropdown Tests
 	
   @22
   Scenario: 22 Prime NG Dropdown
-    Given I am on the Prime NG Dropdown Page
+    Given I am on the Prime NG Home Page
+    When I click the Components Button
+    Then I am redirected to the Prime NG Setup Page
```
dougnoel, line 20: If we have a link directly to the Prime NG Setup page, then we should be going directly there and cut out extraneous steps. We are specifically testing our interaction with the PrimeNG dropdown implementation. We are not testing navigation there.

If we cannot get to the setup page directly with the given URL, then we should delete the URL from that page object.

src/test/java/org/primefaces/PrimeNGHomePage.yml
```diff
@@ -0,0 +1,7 @@
+urls:
+  default: "https://www.primefaces.org/primeng"
+
+elements:
+  components_button:
+    elementType: Button
+    xpath: "//a[@href='/primeng/setup']//span[text()='Components']"
```
dougnoel, line 7: If the url in the PrimeNGSetupPage works, this entire file should be deleted.

src/test/java/org/primefaces/PrimeNGSetupPage.yml
```diff
@@ -0,0 +1,11 @@
...
+elements:
+  dropdown_button:
+    elementType: Button
+    xpath: "//a[@href='/primeng/dropdown']"
+  get_started_header:
+    xpath: "//h1[text()='Get Started']"
+
+
```
dougnoel, line 11: Delete trailing spaces in all files. In yaml files it can cause weird shit to happen but it's just overall unclean.

dougnoel (changes requested):

src/test/java/features/Dropdown Tests.feature
pturchinetz: This is because of that navigation issue where if we go directly it redirects and breaks, We can remove the validation but we do need to navigate from home -> setup -> dropdown

dougnoel (approved):

PR #368, opened 2022-05-27
Fix 367

### LINKED ISSUE #367: Custom element types in projects extending sentinel cause exception
1. use any test project that consumes sentinel and provides custom element types
2. reference an element of a custom element type anywhere in any test

Expected condition: sentinel loads the custom type, finds the element, and performs the step

Error seen: 
```
com.dougnoel.sentinel.exceptions.FileException: <path to page object file, redacted> could not find suitable class for element in this file. Caught

	at com.dougnoel.sentinel.elements.ElementFactory.createElement(ElementFactory.java:101)
	at com.dougnoel.sentinel.pages.Page.lambda$getElement$0(Page.java:77)
	at java.base/java.util.HashMap.computeIfAbsent(HashMap.java:1220)
	at com.dougnoel.sentinel.pages.Page.getElement(Page.java:77)
	at com.dougnoel.sentinel.elements.ElementFunctions.getElement(ElementFunctions.java:34)
```

This is caused by a recent change to the FileManager.getClassPath() method.

sampacos:
Fix #367 
revert breaking change to filemanager.getclasspath() in order to properly fetch custom element type class paths.

### DIFF
--- a/src/main/java/com/dougnoel/sentinel/system/FileManager.java
+++ b/src/main/java/com/dougnoel/sentinel/system/FileManager.java
@@ -100,7 +100,7 @@ public class FileManager {
 	public static String getClassPath(String className) {
 		try {
 			String filePath = findFilePath(className + ".java").getPath();
-			String returnValue = convertPathSeparators(StringUtils.removeEnd(filePath, ".java"));
+			String returnValue = StringUtils.removeEnd(filePath, ".java").replace(File.separator, ".");
 			return StringUtils.substringAfter(returnValue, "java.");
 		} catch (FileException fe) {
 			return null;


pturchinetz (approved):
Seems like a valid change, but can you explain what the issue was in greater detail? Im curious what exactly was the corner case that caused this '.' that we didn't have before

sampacos:
> Seems like a valid change, but can you explain what the issue was in greater detail? Im curious what exactly was the corner case that caused this '.' that we didn't have before

The custom element class lookup requires that the string returned from this method is in the format `com.dougnoel.sentinel.elements.Table`, for example. We construct this successfully by finding the file location and then replacing file separators with `.`. The change that this PR is reverting removed the code that did that replacement.

PR #370, opened 2022-06-01
Add support for the chromium browser.

gusennan:
Adds support for the chromium driver, which is the fully open-source variant of Chrome.


gusennan:
(P.S. this is Nate G. from work)

dougnoel (approved):
I'm good with this as long as it passes the tests and code coverage is above 80%

tyBouch:
Looks like we just need to re-run the build, and then update the branch with master.

gusennan:
I pulled from master.  I was wondering if it's also necessary to install chromium in the continuous integration VM? The error message in the test seemed to indicate it couldn't find chromium. If so, could someone point me to where I would do that?

gusennan:
closing in favor of https://github.com/dougnoel/sentinel/pull/416

PR #372, opened 2022-06-07
371 location

sampacos:
- add location getter on element
- add column location checking method on table
- add verification step for element relative location
- add tests

### DIFF
--- a/src/main/java/com/dougnoel/sentinel/elements/tables/Table.java
+++ b/src/main/java/com/dougnoel/sentinel/elements/tables/Table.java
@@ -810,4 +810,16 @@ public class Table extends Element {
 		return cellValues;
 	}
 
+	/**
+	 * Verifies that the first column header is displayed before (to the left of) the second column header.
+	 * @param column1 String the name of the first column
+	 * @param column2 String the name of the second column
+	 * @return boolean true if the first column header's X coordinate is less than the second column header's X coordinate
+	 */
+	public boolean verifyColumnDisplayOrder(String column1, String column2){
+		WebElement column1HeaderElement = getColumnHeaderElement(column1);
+		WebElement column2HeaderElement = getColumnHeaderElement(column2);
+		return column1HeaderElement.getLocation().x < column2HeaderElement.getLocation().x;
+	}
+
 }
--- a/src/main/java/com/dougnoel/sentinel/steps/TableVerificationSteps.java
+++ b/src/main/java/com/dougnoel/sentinel/steps/TableVerificationSteps.java
@@ -275,4 +275,14 @@ public class TableVerificationSteps {
         }
     }
 
+    @Then("^I verify the (.*?) column in the (.*?) is displayed to the left of the (.*?) column$")
+    public static void verifyColumnDisplayOrder(String column1Name, String tableName, String column2Name){
+        var expectedResult = SentinelStringUtils.format(
+                "Expected the {} column of the {} to be displayed to the left of the {} column.",
+                column1Name, tableName, column2Name);
+        log.trace(expectedResult);
+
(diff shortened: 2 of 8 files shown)


src/main/java/com/dougnoel/sentinel/elements/tables/Table.java
```diff
@@ -810,4 +810,16 @@ private String getCellsValue(List<Integer> indexes, List<String> cells) {
 		return cellValues;
 	}
 
+	/**
+	 * Verifies that the first column header is displayed before (to the left of) the second column header.
+	 * @param column1 String the name of the first column
+	 * @param column2 String the name of the second column
+	 * @return boolean true if the first column header's X coordinate is less than the second column header's X coordinate
+	 */
+	public boolean verifyColumnDisplayOrder(String column1, String column2){
```
tyBouch, line 819: Curious why we're doing this in the table object itself? Is this really used so often?

src/main/java/com/dougnoel/sentinel/steps/TableVerificationSteps.java
```diff
@@ -275,4 +275,14 @@ public static void verifyColumnIsEmpty(String columnName, String tableName, Stri
...
     }
 
+    @Then("^I verify the (.*?) column in the (.*?) is displayed to the left of the (.*?) column$")
+    public static void verifyColumnDisplayOrder(String column1Name, String tableName, String column2Name){
+        var expectedResult = SentinelStringUtils.format(
+                "Expected the {} column of the {} to be displayed to the left of the {} column.",
+                column1Name, tableName, column2Name);
+        log.trace(expectedResult);
+
+        assertTrue(expectedResult, getElementAsTable(tableName).verifyColumnDisplayOrder(column1Name, column2Name));
```
tyBouch, line 285: Yeah just curious if the table implementation is truly better than just doing it in here as it's so specific of a scenario.

tyBouch: Also if we're doing this step we can easily extend this to do a right or left check. Might be useful for later.

src/main/java/com/dougnoel/sentinel/elements/tables/Table.java
sampacos: It doesn't have much to do with how often we're doing it. It lives in Table because the table knows this information about itself. It's in line with the existing design of Table functionality.

tyBouch: Roger. Answers my question.

src/main/java/com/dougnoel/sentinel/steps/TableVerificationSteps.java
tyBouch: ...Point taken!

tyBouch (approved):
Given a look through with questions answered I'm not seeing any issues.

src/main/java/com/dougnoel/sentinel/elements/tables/Table.java
sampacos: For clarity I'll contrast this with the other step in this PR for element location: the reason the location of two elements should be checked in the step is because an element should not have to know information about any other element. To create a method in the Element class that compares one instance of Element to another instance of Element would be an anti-pattern. On the other hand, the columns are a part of the Table, and the Table should know any information related to its columns (and rows, cells, etc.). It makes sense that we can query the Table about the relative position of its columns and it can tell us "yes" or "no". Not the case for any given element in relation to some other element. So that's why I put the code where it is in each instance. Hope that helps clear things up!

PR #374, opened 2022-06-15
Tybou/wait until stored text

tyBouch:
-Modified waitfortext to support value as well for elements that use value for display text (textboxes, etc)
-Modified the random enter text to support true random and sequential unique as well as append/prepend functionality and storing using the element name*
-Added wait until text contains stored text (Might be able to be merged with the other text contains, but I'll need some help with that cucumber regex. Cheat sheet would be fantastic if anyone has a good one)
-Added elements to guinea pig page for testing
-Added test coverage for the new random operation, and modified the old

***Note, we should make this change well known as the step had to be modified to make more sense. We may be able to support the old method as well, but we'll need a unique selector that can exclude "For the" in the before and after match or optionally everything at the end without for the.

If this is desired I can look into it more.**

### DIFF
--- a/src/test/java/features/Text Tests.feature
+++ b/src/test/java/features/Text Tests.feature
@@ -14,7 +14,7 @@ Feature: Text Verifcation Tests
   	Then I verify the username field is empty  
   	When I fill the account information into the username field and the password field
   	  And I enter Bob in the username field
-  	  And I randomly enter text in the password field for the fun of it
+  	  And I randomly enter text in the password field
   	  And I reuse the password field text in the password field
   	Then I verify the username field is not empty
   	  And I verify the username field does not contain the text "foo"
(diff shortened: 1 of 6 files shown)


tyBouch:
Note the quality gate failure is because SonarCloud is presuming the random alphanumeric entry the class used previously is now being used as weak cryptography? We'll either need an approval to bypass it, or switch to secure random number generation. I don't believe secure random number generation is appropriate for this particular use, though, as we don't intend for it to be used as cryptography.

src/test/java/features/Text Tests.feature
```diff
@@ -14,11 +14,30 @@ Feature: Text Verifcation Tests
...
+  	  And I press the browser refresh button
+
+	@text
+	Scenario: I randomly append/prepend/enter text, and verify the entry
+		Given I am on the Guinea Pig Page
+		When I randomly enter test in the Comments Text Area
+			And I click the Send Button
+		Then I wait until the Your Comments Span contains the text entered for the Comments Text Area
+		When I clear the Comments Text Area
+		  And I uniquely enter test in the Comments Text Area
```
sampacos, line 33: the indenting looks off on this scenario compared to the last, as well as on this line specifically. Not sure what's going on there, might be tabs instead of spaces

tyBouch: Done! This particular file was fighting it.

dougnoel (approved):

PR #376, opened 2022-06-15
fix 375

sampacos:
Indices are correct for tables with no proper header row. add tests.

### DIFF
--- a/src/main/java/com/dougnoel/sentinel/elements/tables/Table.java
+++ b/src/main/java/com/dougnoel/sentinel/elements/tables/Table.java
@@ -624,14 +624,16 @@ public class Table extends Element {
 	}
 
 	/**
-	 * Returns true if the cell in the given column and row match the text value given.
+	 * Returns null if the cell in the given column and row match the text value given.
+	 * Otherwise, returns the actual text of the given cell.
 	 *
 	 * @param columnHeader String the name of the column
 	 * @param rowIndex int the index of the row
 	 * @param textToMatch String the text that should be in the specified cell
-	 * @param partialMatch boolean if true, this method returns true if the actual cell text contains the given textToMatch.
-	 *                     if false, this method returns true when the actual cell text exactly matches the given textToMatch.
-	 * @return boolean true if the cell given by the passed column + row contains or exactly matches (depending on partialMatch parameter) the given text, false otherwise.
+	 * @param partialMatch boolean if true, this method returns null if the actual cell text contains the given textToMatch.
+	 *                     if false, this method returns null when the actual cell text exactly matches the given textToMatch.
+	 * @return String null if the text of the cell given by the passed column + row contains or exactly matches (depending on partialMatch parameter) the given text.
+	 * 		   Otherwise, returns the actual text of the given cell.
 	 */
 	public String verifySpecificCellContains(String columnHeader, int rowIndex, String textToMatch, boolean partialMatch) {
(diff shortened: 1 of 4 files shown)


src/main/java/com/dougnoel/sentinel/steps/TableVerificationSteps.java
```diff
@@ -219,30 +220,35 @@ public static void verifyColumnSort(String columnName, String tableName, String
...
+
+    	String resultText = table.verifySpecificCellContains(columnName, rowIndex, textToMatch, partialMatch);
     	
     	var expectedResult = SentinelStringUtils.format(
                 "Expected the cell in the {} row and the {} column of the {} to {}contain the text {}. The element contained the text: {}",
-                SentinelStringUtils.ordinal(rowIndex), columnName, tableName, (negate ? "not " : ""), textToMatch, actualText);
+                SentinelStringUtils.ordinal(rowIndex), columnName, tableName, (negate ? "not " : ""), textToMatch, resultText);
     	log.trace(expectedResult);
+
+        boolean actualResult = resultText == null;
```
tyBouch, line 247: Just curious why verifySpecificCellContains returns a string and not boolean.

src/main/java/com/dougnoel/sentinel/elements/tables/Table.java
```diff
@@ -607,33 +622,32 @@ public boolean verifyAnyColumnCellHas(String columnHeader, String textToMatch) {
 		log.debug("No values in the {} column are equal to {}. False result returned. Turn on trace logging level to see all values found.", columnHeader, textToMatch);
 		return false;
 	}
-	
+
 	/**
 	 * Returns true if the cell in the given column and row match the text value given.
```
tyBouch, line 627: If we're going to be returning a string be sure to update this!

tyBouch (changes requested):

src/main/java/com/dougnoel/sentinel/steps/TableVerificationSteps.java
sampacos: We actually need 2 pieces of information from this method, and we changed to a String return type to provide that info. 

1. does the expected text match the actual text?
2. if it doesn't, what is the actual text? so we can log it 
 
If the returned result result is null, we know the verification was successful so we don't need to log the actual text.

src/test/java/com/dougnoel/sentinel/elements/ElementTests.java
```diff
@@ -216,7 +216,7 @@ public void checkElementBackgroundColorDefaultsWhiteIfAllTransparency() {
     @Test(expected = NoSuchElementException.class)
     public void tableColumnDoesNotExist() {
         BaseSteps.navigateToPage("TablePage");
-        TableVerificationSteps.verifyCellInSpecifiedRow("1", "Not a real column", "example table", "contains", "Bob");
+        TableVerificationSteps.verifyCellInSpecifiedRow("1", "Not a real column", "example table", "","contains", "Bob");
```
dougnoel, line 219: Is this intentional?

sampacos: yup. change in method signature. the empty string is for the "negate" argument of `verifyCellInSpecifiedRow()`. The old code here is actually wrong, since it should be an empty string instead of "contains". Before this branch, there was no option for "contains" vs "has/had".

dougnoel (approved):

PR #378, opened 2022-06-21
Tyler/executable relative paths

tyBouch:
Added support for relative paths with executables

### DIFF
--- a/src/test/java/com/dougnoel/sentinel/system/FileManagerTests.java
+++ b/src/test/java/com/dougnoel/sentinel/system/FileManagerTests.java
@@ -1,6 +1,8 @@
 package com.dougnoel.sentinel.system;
 
+import com.dougnoel.sentinel.exceptions.FileException;
 import com.dougnoel.sentinel.strings.SentinelStringUtils;
+import org.junit.After;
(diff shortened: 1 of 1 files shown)


src/test/java/com/dougnoel/sentinel/system/FileManagerTests.java
```diff
@@ -35,4 +46,47 @@ public void testPathSeparatorReplacement() {
 
         assertTrue(resultMessage, resultPathString.contains(File.separator) && !resultPathString.contains("/"));
     }
+
+    @Test
+    public void localAppDataTranslation() {
```
sampacos, line 51: If this test and the two below run on mac/linux will they fail? Should they be setting the OS to windows?

sampacos: From what I can tell this test class already has a teardown that's resetting the system property to reflect the actual OS of the system. Each of these methods would only need 1 line at the beginning to set the OS property value to "windows" in order to make them run on any platform. But I'm happy to defer to Doug's judgement on this. Just something I thought I should bring up.

Issue #379, opened 2022-06-21
Add ability to use stored text in table's clickAssociatedLinkInTable step


sampacos:
I would be nice to be able to find a row containing stored text, or to click a cell element containing stored text with this step.

PR #380, opened 2022-06-22
379 add ability to use stored values in clickAssociatedLinkInTable table step

sampacos:
address #379  
add ability to use stored values in clickAssociatedLinkInTable table step


tyBouch:
Not seeing anything wrong off the bat. Everything looks in order, so if everything is passing I'm good on my end.

tyBouch (approved):

PR #388, opened 2022-07-15
Adding a .Strip to remove the white spaces when searching the column …

plmlkrz:
Added a .strip() to strip out the white spaces when searching for the column headers.  Also ran the Table Tests.feature that all passed.

### DIFF
--- a/src/main/java/com/dougnoel/sentinel/elements/tables/Table.java
+++ b/src/main/java/com/dougnoel/sentinel/elements/tables/Table.java
@@ -87,7 +87,7 @@ public class Table extends Element {
 		if (headers.isEmpty()) {
 			getOrCreateHeaderElements();
 			for (WebElement header : headerElements) {
-				String headerText = header.getText().replaceAll("[\\t\\n\\r]+"," ");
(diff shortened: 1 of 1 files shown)


src/main/java/com/dougnoel/sentinel/elements/tables/Table.java
```diff
@@ -87,7 +87,7 @@ protected List<String> getOrCreateHeaders() {
 		if (headers.isEmpty()) {
 			getOrCreateHeaderElements();
 			for (WebElement header : headerElements) {
-				String headerText = header.getText().replaceAll("[\\t\\n\\r]+"," ");
+				String headerText = header.getText().replaceAll("[\\t\\n\\r]+"," ").strip();
```
dougnoel, line 90:
```suggestion
				String headerText = header.getText().replaceAll("[\\t\\n\\r]+","");
```

plmlkrz: Attempted to remove the space and this caused many tests to fail.

PR #389, opened 2022-07-15
Updated TableVerificationSteps for the verifyColumnExists method to V…

plmlkrz:
Updated TableVerificationSteps for the verifyColumnExists method to Verify the given table does not contain the given column.  Added new tests in the Table Tests.feature file to test the does not contain option.


sampacos (approved):
I approve the changes but I think the behavior of Table.verifyColumnExists() needs to be reviewed. Currently it allows for partial matches of table headers, which I'm not sure I agree with.

src/main/java/com/dougnoel/sentinel/steps/TableVerificationSteps.java
```diff
@@ -85,20 +85,29 @@ public static void verifyUniqueColumnText(String columnName, String isMultiCells
...
      * <li>I verify the Discography table contains the Album Name column </li>
+     * <li>I verify the Discography table does not contain the Album Name column </li>
      * </ul>
      * @param tableName String name of the table containing the column
      * @param columnName String name of the column to verify
      */
-    @Then("^I verify the (.*?) contains (?:a|the) (.*?) column$")
-    public static void verifyColumnExists(String tableName, String columnName) {
-        assertTrue(getElementAsTable(tableName).verifyColumnExists(columnName));
+    @Then("^I verify the (.*?)( does not)? contains? (?:a|the) (.*?) column$")
```
dougnoel, line 100: Let's modify this to do a has or a contains like other text steps. has is an equal, contains is a contains.

dougnoel (changes requested):

src/main/java/com/dougnoel/sentinel/steps/TableVerificationSteps.java
dougnoel, line 100:
```suggestion
    @Then("^I verify the (.*?)( does not)? (contains?|has|have) (?:a|the) (.*?) column$")
```

sampacos (approved):

dougnoel (approved):

PR #397, opened 2022-07-27
Tyler/file upload sendkeys

tyBouch:
Implemented sendkeys to input elements with file support.
Can perform both multiline and single line uploads and is confirmed functional in both chrome and firefox.


src/main/java/com/dougnoel/sentinel/steps/UploadSteps.java
```diff
@@ -0,0 +1,46 @@
...
+     * <br>--test1:
+     * <br>---fileLocation: test.jpg
+     * <br>--test2:
+     * <br>---fileLocation: C:/testNotInSrc.jpg
+     * </p>
+     * @param elementName String name of the element to screenshot
+     */
+    @When("^I upload the (.*?) files? to the (.*?)$")
+    public static void sendPathsToInputElement(String testDataFilePaths, String elementName) {
+        String[] filePathLocators = testDataFilePaths.split(", ");
```
sampacos, line 37: Having seen this issue before, I think it's best to split on `,` and then strip whitespace on each resulting string.

sampacos (changes requested):

src/main/java/com/dougnoel/sentinel/steps/UploadSteps.java
tyBouch: Done!

sampacos (approved):

PR #405, opened 2022-08-01
Run trim on each comma separated included files in page object yaml

AMehtaSL:


sampacos:
Something seems to be wrong with the build related to sonar analysis, so this can't be merged in yet. Not an issue with your code, to my eyes. I'll contact @dougnoel to get it fixed. Otherwise this looks great. Thanks for contributing!

sampacos:
@AMehtaSL Okay so the build is failing because you are trying to merge from your forked repo. That is unsupported. Please make the same changes in a branch in the original repo and we'll be able to merge your code then. Thanks!

PR #411, opened 2022-08-02
Added step for checking a specific row in a column against a previous…

tyBouch:
…ly stored value


src/main/java/com/dougnoel/sentinel/steps/TableVerificationSteps.java
```diff
@@ -117,7 +118,10 @@ public static void verifyColumnExists(String tableName, String columnName) {
     @Then("^I verify the (.*?) column in the (.*?) contains the same text (?:entered|selected|used) for the (.*)$")
     public static void verifyStoredTextAppearsInColumn(String columnName, String tableName, String key) {
     	var textToMatch = Configuration.toString(key);
-        String errorMessage = SentinelStringUtils.format("Expected the {} column of the {} to contain any cells with the text {}", columnName, tableName, textToMatch);
+        String errorMessage = SentinelStringUtils.format("No previously stored text was found for the \"{}\" key.", key);
```
pturchinetz, line 121: the \"{}\" paradigm you use is probably the best and clearest way to handle logs. I think we should use it in general. You should create a ticket to have someone update the project to follow this standard/start discussion

pturchinetz (approved):

src/main/java/com/dougnoel/sentinel/steps/TableVerificationSteps.java
tyBouch: That or we can modify sentinelStringUtils. Definitely something to bring up and I'll do so.

Especially for this one in particular the key was multi part and it didn't stand out enough on its own being 4 words long.

PR #412, opened 2022-08-02
Tyler/update capable table page

tyBouch:
Added a very simple html test page where the second row can be updated to test both stored column contains, and stored in specific row,col steps


tyBouch:
Used this chance to modify the multi upload test to use element creator so the browser itself performs the necessary escaping to go along with the updatable table page.

sampacos (approved):

PR #414, opened 2022-08-11
Table stored values

david-ondrus:
Add functionality to click and verify stored values in tables

### DIFF
--- a/src/main/java/com/dougnoel/sentinel/steps/TableSteps.java
+++ b/src/main/java/com/dougnoel/sentinel/steps/TableSteps.java
@@ -2,6 +2,7 @@ package com.dougnoel.sentinel.steps;
 
 import static com.dougnoel.sentinel.elements.ElementFunctions.getElementAsTable;
 
+import com.dougnoel.sentinel.elements.tables.Table;
 import com.dougnoel.sentinel.strings.SentinelStringUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.junit.Assert;
@@ -153,19 +154,17 @@ public class TableSteps {
 	 * <p>
 	 * <b>Gherkin Examples:</b>
 	 * <ul>
-	 * <li>I find the Name column in the Project Files Table in row with value entered for the Search Input and click the xpath //i[contains(concat(' ', @class, ' '), ' fa-toggle-on ')]</li>
+	 * <li>I find row in the Project Files Table with value entered for the Search Input and click the xpath //i[contains(concat(' ', @class, ' '), ' fa-toggle-on ')]</li>
 	 * </ul>
-	 * @param columnName String the name of the column to click the header of
 	 * @param tableName String the name of the table element
 	 * @param key String the key used to retrieve the value
 	 * @param xpath String xpath value to element to click
 	 */
-	@When("^I find the (.*?) column in the (.*?) in row with value (?:entered|selected|used) for the (.*?) and click the xpath (.*?)$")
-	public static void clickStoredTextRowContainsXpath(String columnName, String tableName, String key, String xpath) {
+	@When("^I find row in the (.*?) with value (?:entered|selected|used) for the (.*?) and click the xpath (.*?)$")
+	public static void clickStoredTextRowContainsXpath(String tableName, String key, String xpath) {
+		By locator = By.xpath(xpath);
+		Table table = getElementAsTable(tableName);
 		var textToMatch = Configuration.toString(key);
-		String errorMessage = SentinelStringUtils.format("No previously stored text was found for the \"{}\" key.", key);
-		Assert.assertNotNull(errorMessage, textToMatch);
-		var rowIndex = getElementAsTable(tableName).getAllCellDataForColumn(columnName).indexOf(textToMatch);
-		getElementAsTable(tableName).getElementInRowThatContains(rowIndex, By.xpath(xpath)).click();
+		table.clickElementInRowThatContains(textToMatch, locator);
 	}
 }
--- a/src/main/java/com/dougnoel/sentinel/steps/TableSteps.java
+++ b/src/main/java/com/dougnoel/sentinel/steps/TableSteps.java
@@ -145,4 +145,25 @@ public class TableSteps {
     public static void clickColumnHeaderToSort(String columnName, String tableName) {
     	getElementAsTable(tableName).clickColumnHeader(columnName);
     }
+
+	/**
+	 * Clicks the xpath of element in row which have stored value in column name
+	 * <p>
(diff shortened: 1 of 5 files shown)


src/main/java/com/dougnoel/sentinel/steps/TableSteps.java
```diff
@@ -145,4 +147,25 @@ public static void storeTableValues(int pageNumber, String tableName) {
...
+	 * @param columnName String the name of the column to click the header of
+	 * @param tableName String the name of the table element
+	 * @param key String the key used to retrieve the value
+	 * @param xpath String xpath value to element to click
+	 */
+	@When("^I find the (.*?) column in the (.*?) in row with value (?:entered|selected|used) for the (.*?) and click the xpath (.*?)$")
+	public static void clickStoredTextRowContainsXpath(String columnName, String tableName, String key, String xpath) {
+		var textToMatch = Configuration.toString(key);
+		String errorMessage = SentinelStringUtils.format("No previously stored text was found for the \"{}\" key.", key);
+		Assert.assertNotNull(errorMessage, textToMatch);
```
dougnoel, line 167: This should be done as `assertNotNull(errorMessage, testToMatch);` And include should be
`import static org.junit.Assert.assertNotNull;` However, this is not a test so this assertion shouldn't be used here. (See my other comment.)

src/main/java/com/dougnoel/sentinel/steps/TableSteps.java
```diff
@@ -145,4 +147,25 @@ public static void storeTableValues(int pageNumber, String tableName) {
...
+		var rowIndex = getElementAsTable(tableName).getAllCellDataForColumn(columnName).indexOf(textToMatch);
+		getElementAsTable(tableName).getElementInRowThatContains(rowIndex, By.xpath(xpath)).click();
+	}
```
dougnoel, line 170: The null assert should be handled as a try/catch. Otherwise we are doing an evaluation every time which slows down the program.

dougnoel (changes requested):

pturchinetz (dismissed):

src/main/java/com/dougnoel/sentinel/steps/TableSteps.java
```diff
@@ -2,7 +2,10 @@
 
 import static com.dougnoel.sentinel.elements.ElementFunctions.getElementAsTable;
 
+import com.dougnoel.sentinel.elements.tables.Table;
+import com.dougnoel.sentinel.strings.SentinelStringUtils;
 import org.apache.commons.lang3.StringUtils;
+import org.junit.Assert;
```
sampacos, line 8: remove unused import

src/main/java/com/dougnoel/sentinel/steps/TableSteps.java
```diff
@@ -2,7 +2,10 @@
 
 import static com.dougnoel.sentinel.elements.ElementFunctions.getElementAsTable;
 
+import com.dougnoel.sentinel.elements.tables.Table;
+import com.dougnoel.sentinel.strings.SentinelStringUtils;
```
sampacos, line 6: same thing here. remove unused import

src/main/java/com/dougnoel/sentinel/steps/TableSteps.java
```diff
@@ -145,4 +148,23 @@ public static void storeTableValues(int pageNumber, String tableName) {
...
+	 * <b>Gherkin Examples:</b>
+	 * <ul>
+	 * <li>I find row in the Project Files Table with value entered for the Search Input and click the xpath //i[contains(concat(' ', @class, ' '), ' fa-toggle-on ')]</li>
+	 * </ul>
+	 * @param tableName String the name of the table element
+	 * @param key String the key used to retrieve the value
+	 * @param xpath String xpath value to element to click
+	 */
+	@When("^I find row in the (.*?) with value (?:entered|selected|used) for the (.*?) and click the xpath (.*?)$")
+	public static void clickStoredTextRowContainsXpath(String tableName, String key, String xpath) {
```
sampacos, line 164: Is this functionality not already covered by `TableSteps.clickAssociatedLinkInTable`? How is this different?

src/main/java/com/dougnoel/sentinel/steps/TableVerificationSteps.java
```diff
@@ -321,4 +323,23 @@ public static void verifyColumnDisplayOrder(String column1Name, String tableName
...
+     * <b>Gherkin Examples:</b>
+     * <ul>
+     * <li>I verify row in the Project Files Table with value selected for the Search Input contains the xpath //i[contains(concat(' ', @class, ' '), ' fa-toggle-on ')]</li>
+     * </ul>
+     * @param tableName String the name of the table element
+     * @param key String the key used to retrieve the value
+     * @param xpath String xpath value to element to click
+     */
+    @Then("^I verify row in the (.*?) with value (?:entered|selected|used) for the (.*?) contains the xpath (.*?)$")
+    public static void verifyStoredTextRowContainsXpath(String tableName, String key, String xpath) {
```
sampacos, line 338: this doesn't seem to follow the assertion / verification pattern we have in most of the other TableVerificationSteps. I'd like to see us create a table method that returns a boolean, and then keep that verification pattern going where we use `assertTrue(errorMessage, table, table -> table.verificationMethod)`  I'd also like to see the negative outcome covered, in the case that we want to verify a certain row does not contain an element matched by an xpath.

src/main/java/com/dougnoel/sentinel/steps/TableVerificationSteps.java
```diff
@@ -321,4 +323,23 @@ public static void verifyColumnDisplayOrder(String column1Name, String tableName
...
+     * <p>
+     * <b>Gherkin Examples:</b>
+     * <ul>
+     * <li>I verify row in the Project Files Table with value selected for the Search Input contains the xpath //i[contains(concat(' ', @class, ' '), ' fa-toggle-on ')]</li>
+     * </ul>
+     * @param tableName String the name of the table element
+     * @param key String the key used to retrieve the value
+     * @param xpath String xpath value to element to click
+     */
+    @Then("^I verify row in the (.*?) with value (?:entered|selected|used) for the (.*?) contains the xpath (.*?)$")
```
sampacos, line 337: I think this would read better if we added a "the" in front of "row" and "value". So `I verify the row in the _ with the value ...`. If you want to make them optional, that works too.

sampacos (changes requested):

pturchinetz (approved):

sampacos (approved):

PR #416, opened 2022-08-21
Add support for the chromium browser.

gusennan:
Adds support for the open source chromium browser.

We install chromium into the GitHub workflow using chocolately (viz.
choco install chromium). By default chromedriver looks for chrome to be
installed in a standard location, which chrome is, so we override the
location for which to look for chome.exe, which is the chromium binary.

Add a basic sanity unit test that ensures the chromium browser can be
launched and made fullscreen.

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -41,6 +41,8 @@ jobs:
       run: |
        chmod +x ./src/test/resources/scripts/InstallChrome.sh
         ./src/test/resources/scripts/InstallChrome.sh
+    - name: Install Chromium
+      run: choco install --no-progress -y chromium
     - name: Build with Maven and Analyze with Sonar
       env:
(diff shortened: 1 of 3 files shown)


dougnoel (changes requested):
This looks good, but I need to see the code coverage report. The easiest way to do this is to merge your code into a branch on here and then create a pull request from that branch. If you do not have permissions, let me know and I will give them to you.

gusennan:
I definitely don’t currently have permissions to create branches in this repo.On Sep 28, 2022, at 1:33 PM, Doug Noël ***@***.***> wrote:﻿
@dougnoel requested changes on this pull request.

This looks good, but I need to see the code coverage report. The easiest way to do this is to merge your code into a branch on here and then create a pull request from that branch. If you do not have permissions, let me know and I will give them to you.

PR #422, opened 2022-08-25
Added the new textbox element and updated the page object

tyBouch:


docs/test/textbox.html
```diff
@@ -5,25 +5,27 @@
...
-		  <input type="checkbox" id="vehicle3" name="vehicle3" value="Boat" checked>
-		  <label for="vehicle3"> I have a boat</label><br><br>
-		  <label for="fname">First name:</label>
-		  <input type="text" id="fname" name="fname" disabled><br><br>
-		  <label for="lname">Last name:</label>
-		  <input type="text" id="lname" name="lname"><br><br>
-		  <input type="submit" value="Submit">
-		  <input type="text" id="hiddenbox" class="hidden">
-		  <input type="text" id="notdisplayedbox" class="notdisplayed">
+			<input type="checkbox" id="vehicle1" name="vehicle1" value="Bike">
```
pturchinetz, line 14: Is the tabbing off or is this just a visual thing here?

pturchinetz (commented):
looks fine, the formatting seems off though, i could be wrong

docs/test/textbox.html
sampacos: looks like it's been changed to be more consistent with the rest of the document. Don't think it's a problem so I'm fine with it.

sampacos (approved):

docs/test/textbox.html
tyBouch: Yeah it wasn't quite tabbed the right way before. Same with the page object which used non standard tabbing before.

Issue #386, opened 2022-07-15
Add Support for Excel


plmlkrz:
Need the ability to Verify that data is present in an Excel File.  

Immediate need: Verify data is in the excel and verify the sheets are created.
Verify the formatting of the excel
Verify that data in row1 column 1 has the correct data.

Future needs: 
Grab data within the Excel, store it in memory  and use it in other cucumber steps.

dougnoel:
1. Create Test Object Abstract Class
2. Inherit Element from that object
3. Inherit Excel File from the Test Object class
4. Handle the creation of an Excel object
5. Change the implementation of a table object to use composition instead of inheritance to allow the use of Excel files
6. Implement an Excel file.
7. Implement an Excel table as a range in a sheet in an Excel File
8. Allow the identification of a cell inside an excel table for validation

PR #423, opened 2022-08-26
386 support csv

sampacos:
Addresses #386 partially

- add CSV file verification and editing. 
- add concept of abstract TestFile. 
- add tests for new code. 
- add utility method for strings. 
- add steps for dealing with CSV files. 
- add 2 new CSV files to use for testing. 
- add feature file to test step definitions. 
- add unit tests for constructors and other CSV methods. 
- update configuration testdata to be more stable and be able to replace testdata names with spaces, like with elements.


src/main/java/com/dougnoel/sentinel/files/CsvFile.java
```diff
@@ -0,0 +1,334 @@
...
+import java.util.Objects;
+
+import com.dougnoel.sentinel.exceptions.FileException;
+import com.dougnoel.sentinel.strings.SentinelStringUtils;
+import org.apache.commons.csv.CSVFormat;
+import org.apache.commons.csv.CSVParser;
+import org.apache.commons.csv.CSVPrinter;
+import org.apache.commons.lang3.StringUtils;
+
+public class CsvFile extends TestFile{
```
pturchinetz, line 20: I'd suggest adding functionality to switch between worksheets (tabs)

src/main/java/com/dougnoel/sentinel/steps/CsvSteps.java
```diff
@@ -0,0 +1,142 @@
...
+     * @param matchType String whether we are doing an exact match or a partial match.
+     * @param textToMatch String the text to look for in the cell.
+     * @param column String column the name of the column, or an ordinal (1st, 2nd, 25th, etc.).
+     * @param rowNum String the row number. Can be "la" to specify the last row, or an integer.
+     */
+    @Then("^I verify the (?:CSV|csv) file( do(?:es)? not)? (has|have|contains?) the value (.*) in the (.*) column and the (la|\\d+)(?:st|nd|rd|th) row$")
+    public static void verifyCsvCellHasValue(String assertion, String matchType, String textToMatch, String column, String rowNum) {
+        CsvFile file = (CsvFile) FileManager.getCurrentTestFile();
+        boolean negate = !StringUtils.isEmpty(assertion);
+        int rowIndex = rowNum.equals("la") ? file.getNumberOfRows() : Integer.parseInt(rowNum);
```
pturchinetz, line 82: I could be blind but where do you handle the "la" case? I couldn't find it.

src/main/java/com/dougnoel/sentinel/steps/UploadSteps.java
```diff
@@ -43,4 +45,9 @@ public static void sendPathsToInputElement(String testDataFilePaths, String elem
 
         getElement(elementName).sendFilePaths(filesToUpload);
     }
+
```
pturchinetz, line 48: Function header

src/main/java/com/dougnoel/sentinel/system/DownloadManager.java
```diff
@@ -384,4 +389,23 @@ public static void setDownloadDirectory(String downloadDirectory) {
     public static void clearDownloadDirectory() throws IOException {
         FileUtils.cleanDirectory(new File(downloadDirectory));
     }
+
+    /**
+     * Sets the path of the most recently-downloaded file to the current download directory + the given filename.
+     * @param filename String name of the file downloaded.
+     */
+    private static void setMostRecentDownloadPath(String filename){
```
pturchinetz, line 397: Shame we aren't in C# and can't use {get; set; }

src/main/java/com/dougnoel/sentinel/files/CsvFile.java
sampacos: Csv files don't have worksheets. That's an excel thing that will have to be implemented later.

src/main/java/com/dougnoel/sentinel/files/CsvFile.java
```diff
@@ -0,0 +1,334 @@
...
+     * @throws FileNotFoundException in the case that a file does not exist at that path.
+     */
+    public CsvFile(Path pathToFile, CSVFormat format, int numberOfHeaderRows) throws FileNotFoundException {
+        super(pathToFile);
+        csvFormat = format;
+        numHeaderRows = numberOfHeaderRows;
+    }
+
+    @Override
+    public boolean equals(Object o) {
```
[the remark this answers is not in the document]
sampacos: 1. this is an auto-generated method created by IntelliJ. It's recommended that any non-static class implements an equals() method so it returns accurately based on the fields and such on the object.
2. If I understand java correctly, the first comparison (line 96) does a check based on the object in the java heap. It will return true if the compiler's reference in the heap to both objects is the same. Not sure what you mean by loading issues. It's strictly checking whether the object passed exists in the same location in memory as the current object. if so, that means they're the exact same object.
3. the last line returns true or false, depending on whether the object passed in has values in those fields that match the current instance's values. basically, it will only reach that line if it's determined that the object passed in is the same type (CsvFile) but not the same memory location, and if those values match the current instance's, then we can say the two objects are equal.
hope that helps.

src/main/java/com/dougnoel/sentinel/files/CsvFile.java
pturchinetz: Right! Sorry, got ahead of myself there.

src/main/java/com/dougnoel/sentinel/files/CsvFile.java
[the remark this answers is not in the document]
pturchinetz: Oh, I had no idea it was auto-generated. Lets keep it the way it is then. I just read up on java == for objects, I think you're right about that. That explains the last line too, I thought we were just comparing counts instead of the whole row. Thanks!

src/main/java/com/dougnoel/sentinel/files/CsvFile.java
```diff
@@ -0,0 +1,334 @@
...
+            log.trace("IOException caught while parsing column headers.");
+            return Collections.emptyList();
+        }
+    }
+
+    /**
+     * Sets the whole file's contents to the passed newFileContents values, where each list of lists is a row.
+     * @param newFileContents List&lt;List&lt;String&gt;&gt; All contents of the file to write.
+     */
+    public void setFileContents(List<List<String>> newFileContents){
```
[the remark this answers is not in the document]
sampacos: same thing with the worksheets. not implemented here because CSV files do not have them.

src/main/java/com/dougnoel/sentinel/files/CsvFile.java
```diff
@@ -0,0 +1,334 @@
...
+     * Returns the number of rows in the CSV file.
+     * @return int the number of rows in the CSV file.
+     */
+    public int getNumberOfRows(){
+        try(var parser = getParser()){
+            return parser.getRecords().size();
+        }
+        catch(IOException ioe){
+            log.trace(SentinelStringUtils.format("IOException caught while parsing CSV file {}.", toPath()));
+            return -1;
```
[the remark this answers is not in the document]
sampacos: Without any further info I'm not sure if that would apply here. Any step that takes "last" as a row input gets the result of this method and subtracts 1, then puts that value into the validation method. the input -1 is not a valid input to any of those methods.

src/main/java/com/dougnoel/sentinel/files/CsvFile.java
```diff
@@ -0,0 +1,334 @@
...
+        CsvFile csvFile = (CsvFile) o;
+        return numHeaderRows == csvFile.numHeaderRows && Objects.equals(csvFormat, csvFile.csvFormat);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(super.hashCode(), csvFormat, numHeaderRows);
+    }
+
+    private CSVParser getParser() throws IOException {
```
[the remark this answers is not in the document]
sampacos: This is mainly due to the nature of the CSV library I'm using. The parser is essentially an iterator over the file, which cannot go backwards. That means if we kept the same parser object between method calls it could miss parts of the file. If method call 1 was to read the 3rd row, and method call 2 was to get all contents of the file, method call 2 would not return the first 3 rows. We have to make a new parser for each method call. This is also advantageous because it means every time we read from the file, we have the most up-to-date contents, and we don't run into access conflicts. when we're between reads, we are not technically accessing the file. So other processes are free to change it.

src/main/java/com/dougnoel/sentinel/files/CsvFile.java
```diff
@@ -0,0 +1,334 @@
...
+        int columnIndex = getColumnIndex(columnHeader);
+        setAllCellsInColumn(columnIndex, newValue);
+    }
+
+    /**
+     * Sets all the cells in the given column to the given value.
+     * @param columnIndex int index of the column, starting at 1.
+     * @param newValue String value to set each cell to.
+     */
+    public void setAllCellsInColumn(int columnIndex, String newValue){
```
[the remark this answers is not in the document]
sampacos: You're right but this PR is large enough and I had to call it at some point. I hope the rest of this file is good enough to let someone add that code easily when they need it.

src/main/java/com/dougnoel/sentinel/steps/CsvSteps.java
```diff
@@ -0,0 +1,142 @@
...
+import io.cucumber.java.en.Then;
+import io.cucumber.java.en.When;
+import org.apache.commons.lang3.StringUtils;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+import java.io.FileNotFoundException;
+import java.nio.file.Path;
+
+import static org.junit.Assert.*;
```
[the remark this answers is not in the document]
sampacos: To my eye we've only got at most 2 from a class in the above imports. Is that enough to warrant using the *? Personally I never really notice import sections because the IDE collapses them, but if we have some standard I'm happy to abide by it.

src/main/java/com/dougnoel/sentinel/steps/CsvSteps.java
sampacos: It's in line 82: `rowNum.equals("la")`

src/main/java/com/dougnoel/sentinel/system/DownloadManager.java
sampacos: agreed

src/main/java/com/dougnoel/sentinel/files/TestFile.java
```diff
@@ -0,0 +1,35 @@
...
+import java.nio.file.Path;
+
+public abstract class TestFile extends File{
+    protected static final Logger log = LogManager.getLogger(TestFile.class.getName()); // Create a logger.
+
+    /**
+     * Creates a TestFile from the DownloadManager's latest download and immediately checks that it exists. If it doesn't, an exception is thrown.
+     * @throws FileNotFoundException in the case that a file is not found at the path given by the DownloadManager,
+     * or if a file has not been downloaded in this test session.
+     */
```
dougnoel, line 19:
```suggestion
    /**
     * Creates a TestFile from the DownloadManager's latest download and immediately checks that it exists. If it doesn't, an exception is thrown.
     * <br />
     * @throws FileNotFoundException in the case that a file is not found at the path given by the DownloadManager,
     * or if a file has not been downloaded in this test session.
     */
```

src/main/java/com/dougnoel/sentinel/files/CsvFile.java
```diff
@@ -0,0 +1,334 @@
...
+
+    /**
+     * Create default CSV file from most recently-downloaded path.
+     * @throws FileNotFoundException in the case that a file does not exist at that path.
+     */
+    public CsvFile() throws FileNotFoundException {
+        super();
+        csvFormat = CSVFormat.DEFAULT;
+        numHeaderRows = 1;
+    }
```
dougnoel, line 33: See if you can consolidate to reuse CsvFile(int numHeaderRows)

src/main/java/com/dougnoel/sentinel/files/CsvFile.java
```diff
@@ -0,0 +1,334 @@
...
+        }
+    }
+
+    /**
+     * Returns a List&lt;String&gt; consisting of each value in a given row.
+     * The row index starts at 1 and factors out the header rows, so that calling this method with rowIndex=1 gets the first row with actual data.
+     * @param rowIndex int index of the first row with real data, starting at 1, taking into consideration row headers.
+     * @return List&lt;String&gt; consisting of each value in a given row.
+     */
+    public List<String> getRowData(int rowIndex){
```
dougnoel, line 132: Let's edit all these to "read" so it's clear we are reading the file every time.
```suggestion
    public List<String> readRowData(int rowIndex){
```

src/main/java/com/dougnoel/sentinel/files/CsvFile.java
```diff
@@ -0,0 +1,334 @@
...
+    public int getColumnIndex(String columnHeader){
+        return getHeaders().indexOf(columnHeader) + 1;
+    }
+
+    /**
+     * Sets all the cells in the given column to the given value.
+     * @param columnHeader String name of the column.
+     * @param newValue String value to set each cell to.
+     */
+    public void setAllCellsInColumn(String columnHeader, String newValue) {
```
dougnoel, line 234:
```suggestion
    public void writeAllCellsInColumn(String columnHeader, String newValue) {
```

src/main/java/com/dougnoel/sentinel/files/CsvFile.java
dougnoel, line 244:
```suggestion
    public void writeAllCellsInColumn(int columnIndex, String newValue){
```

src/main/java/com/dougnoel/sentinel/steps/CsvSteps.java
```diff
@@ -0,0 +1,177 @@
...
+     * <b>Gherkin Examples:</b>
+     * <ul>
+     *     <li>I open a CSV file with 3 header rows</li>
+     *     <li>I open a csv file with 1 header row</li>
+     * </ul>
+     *
+     * @param numberOfHeaderRows int the number of header rows in the CSV file to test. Number of header rows must be declared in order to properly process the file.
+     * @throws FileNotFoundException In the case that the file is not found in the location that the DownloadManager specifies.
+     */
+    @When("^I open a (?:CSV|csv) file with (\\d+) header rows?$")
```
dougnoel, line 35:
```suggestion
    @When("^I open the last downloaded (?:CSV|csv) file with (\\d+) header rows?$")
```

src/main/java/com/dougnoel/sentinel/strings/SentinelStringUtils.java
```diff
@@ -57,4 +58,25 @@ public static String ordinal(int i) {
...
+	 * <b>Examples:</b>
+	 * <ul>
+	 *<li>parseLeadingInt("12th") = 12</li>
+	 *<li>parseLeadingInt("1t2h3") = 1</li>
+	 *<li>parseLeadingInt("ten") throws Exception</li>
+	 *</ul>
+	 * @param stringStartingWithInteger String a string with leading integers.
+	 * @return int the integers in the passed string up until the point of the first non-numeric character.
+	 */
+	public static int parseLeadingInt(String stringStartingWithInteger){
```
dougnoel, line 72:
```suggestion
	public static int parseOrdinal(String stringStartingWithInteger){
```

src/test/java/com/dougnoel/sentinel/files/CsvFileTests.java
```diff
@@ -0,0 +1,60 @@
...
+        var filePath = Path.of(filePathOneHeader);
+        CsvFile file1 = new CsvFile(filePath);
+        assertNotNull("Expected new CsvFile to be created.", file1);
+        CsvFile file2 = new CsvFile(filePath, CSVFormat.DEFAULT);
+        assertNotNull("Expected new CsvFile to be created.", file2);
+        CsvFile file3 = new CsvFile(filePath, CSVFormat.EXCEL, 0);
+        assertNotNull("Expected new CsvFile to be created.", file3);
+        assertEquals("File 1 should equal file 2.", file1, file2);
+        assertNotEquals("File 2 should not equal file 3.", file2, file3);
+    }
```
dougnoel, line 33: One assert per test.

dougnoel (changes requested):

src/main/java/com/dougnoel/sentinel/configurations/Configuration.java
```diff
@@ -449,24 +449,26 @@ public static Map <String,String> getElement(String elementName, String pageName
 	 * @return String the value of the given key in the given object
 	 */
 	public static String getTestdataValue(String testdataObjectName, String testdataObjectKey){
+		String normalizedTestdataObjectName = testdataObjectName.replaceAll("\\s+", "_");
+		String normalizedTestdataObjectKey = testdataObjectKey.replaceAll("\\s+", "_");
```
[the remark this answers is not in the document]
sampacos: new issue, #428

src/main/java/com/dougnoel/sentinel/files/CsvFile.java
```diff
@@ -0,0 +1,334 @@
...
+    /**
+     * Create a CSV file configured with the given number of header rows from the CSV file at the most recently downloaded path.
+     * @param numberOfHeaderRows int number of header rows in the CSV file.
+     * @throws FileNotFoundException in the case that a file does not exist at that path.
+     */
+    public CsvFile(int numberOfHeaderRows) throws FileNotFoundException {
+        super();
+        csvFormat = CSVFormat.DEFAULT;
+        numHeaderRows = numberOfHeaderRows;
+    }
```
dougnoel, line 68: This constructor isn't exercised. I'd like to see it removed if we aren't using it now or have a unit test covering it so someone trying to use it later isn't surprised when the code doesn't work.

src/main/java/com/dougnoel/sentinel/files/TestFile.java
```diff
@@ -0,0 +1,33 @@
...
+public abstract class TestFile extends File{
+    /**
+     * Creates a TestFile from the DownloadManager's latest download and immediately checks that it exists. If it doesn't, an exception is thrown.
+     *
+     * @throws FileNotFoundException in the case that a file is not found at the path given by the DownloadManager,
+     * or if a file has not been downloaded in this test session.
+     */
+    protected TestFile() throws FileNotFoundException {
+        this(DownloadManager.getMostRecentDownloadPath());
+    }
```
dougnoel, line 19: This constructor isn't used. I'd like to see a test for this even if it's just injecting the test file into the correct value in the Download Manager.

dougnoel: This will probably be covered if you do this in the CsVFile constructor.

src/main/java/com/dougnoel/sentinel/files/CsvFile.java
```diff
@@ -0,0 +1,334 @@
...
+    private final CSVFormat csvFormat;
+    private final int numHeaderRows;
+
+    /**
+     * Create default CSV file from most recently-downloaded path.
+     * @throws FileNotFoundException in the case that a file does not exist at that path.
+     */
+    public CsvFile() throws FileNotFoundException {
+        this(1);
+    }
```
dougnoel, line 34: This constructor isn't used. I'd like to see a test for this even if it's just injecting the test file into the correct value in the Download Manager.

dougnoel (approved):

dougnoel (approved):

Issue #430, opened 2022-09-26
Add table method for waiting for a specific cell to have a value


sampacos:
It would be nice to be able to say "I wait for the cell ... to contain the text XYZ", to test asynchronous tasks that update data in the table.

PR #431, opened 2022-09-27
430 table wait

sampacos:
Address #430 
Add Table method for waiting a given amount of time for a cell to have or not have given text. Add step definition to use that method. Add unit tests.

### DIFF
--- a/src/main/java/com/dougnoel/sentinel/steps/TableVerificationSteps.java
+++ b/src/main/java/com/dougnoel/sentinel/steps/TableVerificationSteps.java
@@ -2,6 +2,8 @@ package com.dougnoel.sentinel.steps;
 
 import static com.dougnoel.sentinel.elements.ElementFunctions.getElementAsTable;
 import static com.dougnoel.sentinel.assertions.TableAssert.*;
+
+import com.dougnoel.sentinel.configurations.Time;
 import org.junit.Assert;
 
 import com.dougnoel.sentinel.elements.tables.Table;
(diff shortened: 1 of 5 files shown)


src/main/java/com/dougnoel/sentinel/steps/TableVerificationSteps.java
```diff
@@ -350,4 +352,33 @@ public static void verifyStoredTextRowContainsXpath(String tableName, String key
...
+     * </ul>
+     * @param numberOfSecondsToWait String number of seconds to wait. Must be an integer. Recommended minimum value: 2 seconds.
+     * @param rowNum String row of the cell to check.
+     * @param columnName String column of the cell to check.
+     * @param tableName String name of the table.
+     * @param assertion String if null is passed, looks for match(es), if any strong value is passed, looks for the value to not exist.
+     * @param matchType String whether we are doing an exact match or a partial match.
+     * @param textToMatch String the text to look for in the column.
+     */
+    @Then("^I wait up to (\\d{1,3}) seconds for the cell in the (la|\\d+)(?:st|nd|rd|th) row and the (.*) column of the (.*?) to( not)? (has|have|contains?) the text (.*?)$")
```
dougnoel, line 371: Is there a reason you want to wait a different amount of time that the default amount set by the timeout property?

dougnoel (changes requested):

src/main/java/com/dougnoel/sentinel/steps/TableVerificationSteps.java
sampacos: For things we expect to take over a minute, I think it makes sense to explicitly state in the step. But if you disagree I could make this use the default timeout and just up that to whatever is necessary for the test.

dougnoel (approved):

PR #443, opened 2022-10-28
Adding xPathToBy and xPathToBy(int ordinal)

plmlkrz:
Adding xPathToBy and xPathToBy(int ordinal) to be used for finding an xpath in a table.


src/main/java/com/dougnoel/sentinel/elements/Element.java
```diff
@@ -952,4 +952,25 @@ public Point getLocation(){
 		return element().getLocation();
 	}
 
+	/**
+	 * Returns the stored XPath for this element.
+	 * Returns null if no XPath exists in the page object.
+	 *
+	 * @return the xPath for the element listed in the page object
+	 */
+	public By xPathToBy() {
```
pturchinetz, line 961: See if we can make these private/protected

plmlkrz: Private and Protected created an error  on this By enterTextLocator = getElement(elementName).xPathToBy(ordinalRow); It was suggested to make it public

dougnoel (approved):

PR #444, opened 2022-10-28
Adding replaceAll("[\t\n\r]+"," ") so that it can handle a column hea…

plmlkrz:
Adding replaceAll("[\t\n\r]+"," ") so that it can handle a column header with a new line.


pturchinetz (approved):
id post on the QA channel and tell people to run after incase their column headers have the stuff youre stripping out

Issue #453, opened 2022-11-10
Unable to check that a select element has an empty selection


sampacos:
We need to be able to check that a select element has a blank or empty current selection. This is not possible with the current gherkin.

PR #455, opened 2022-11-10
453 empty select and new download verification step

sampacos:
Address #453 
- update select element to override getText() so it can be used by text
 verification empty method. add download verification step for checking contents of zip. add use of empty step on dropdown.
- add locators to page yaml. add postive and negative test in feature. add error handling case in download manager tests. update message in step.


src/main/java/com/dougnoel/sentinel/steps/DownloadVerificationSteps.java
```diff
@@ -65,4 +70,28 @@ public static void verifyFileWithFilenameDownloaded(String expectedFilename) thr
 	public static void clearDownloadsFolder() throws IOException {
 		DownloadManager.clearDownloadDirectory();
 	}
+
+	/**
+	 * Looks inside the most recently downloaded zip file and asserts that it does or does not contain a file with the given extension in the top-level directory.
+	 * @param assertion String " does not" for a negative check, otherwise positive check
+	 * @param expectedFileType String the file extension to check for.
+	 */
+	@Then("^I verify the most recently downloaded zip file( does not)? contains? a file with the extension (.*?)$")
```
pturchinetz, line 79: Nothing wrong with the function, but it seems like an odd case to include. My instinct would be to have a function to check if a specific file is present inside a zip, not any extension type.

pturchinetz: Like you could check for anything with an extension with 
"I verify the most recently downloaded zip file does not contains a file with the name *.exe"

sampacos: Fair but that's not my use case. I don't care what the name is, and implementing regex / wildcard behavior like that seems like overkill. My opinion is if we need it later we can add it.

src/main/java/com/dougnoel/sentinel/elements/dropdowns/SelectElement.java
```diff
@@ -91,6 +91,15 @@ public String getText(int index){
...
     }
+
+    /**
+     * Gets the text of the first item currently selected.
+     * @return String the text value of the selected option
+     */
+    @Override
+    public String getText(){
+        return getSelectedText();
+    }
```
dougnoel, line 102: Add @deprecated to getSelectedText() if we don't need it anymore and we will remove it in a future version.

src/main/java/com/dougnoel/sentinel/steps/DownloadVerificationSteps.java
```diff
@@ -65,4 +70,28 @@ public static void verifyFileWithFilenameDownloaded(String expectedFilename) thr
...
+		boolean result = false;
+		boolean negate = !StringUtils.isEmpty(assertion);
+		try(var zip = new ZipFile(mostRecentFile)){
+			fileContent = zip.stream().map(ZipEntry::getName).collect(Collectors.toList());
+			result = fileContent.stream().anyMatch(fileName -> StringUtils.endsWith(fileName, expectedFileType));
+		}
+		catch(IOException ioe){
+			String message = SentinelStringUtils.format("Unable to open most recently downloaded file as zip file. Most recently downloaded file path {}", mostRecentFile.getPath());
+			throw new IOException(message, ioe);
+		}
```
dougnoel, line 92: I'd prefer this functionality was in the FileManager. Same reasons I was talking about yesterday.

And why do we only check the first directory level in the file? That seems unnecessarily confusing for people using it.

src/main/java/com/dougnoel/sentinel/steps/DownloadVerificationSteps.java
```diff
@@ -65,4 +70,28 @@ public static void verifyFileWithFilenameDownloaded(String expectedFilename) thr
...
+		assertEquals(SentinelStringUtils.format("Expected zip file to {}contain a file with extension {}. Zip file location: {} Files in zip: {}",
+						(negate ? "not ": ""), expectedFileType, mostRecentFile, StringUtils.join(fileContent, ", ")),
```
dougnoel, line 94: This error message needs to be very clear if we aren't searching subdirectories in the zip.

src/test/java/features/327 Download Verification Steps.feature
```diff
@@ -25,4 +25,11 @@ Feature: Download Verification Steps
     When I navigate to the Upload Page
       And I re-upload the most recently downloaded file to the choose file button
     Then I verify the file counter with the attribute data-filecount has the value 1
-      And I verify the upload result 1 with the attribute data-filename has the value TestPDF.pdf
\ No newline at end of file
+      And I verify the upload result 1 with the attribute data-filename has the value TestPDF.pdf
```
dougnoel, line 28: This seems confusingly worded. Doesn't need to change now, but something to discuss for the future.

dougnoel (changes requested):

src/main/java/com/dougnoel/sentinel/steps/DownloadVerificationSteps.java
dougnoel: I'm ok with that. Filename is a different use case we can implement when needed.

src/main/java/com/dougnoel/sentinel/steps/DownloadVerificationSteps.java
sampacos: The file enumeration actually does find subdirectories (except nested zips) but that is clear in the method description.

src/main/java/com/dougnoel/sentinel/steps/DownloadVerificationSteps.java
sampacos: Moved it to it's own file class. And it actually does find files in all folders besides nested zips. And that's documented on method descriptions.

src/main/java/com/dougnoel/sentinel/elements/dropdowns/SelectElement.java
sampacos: Removed it altogether and refactored a chunk of Select text verification as a result.

src/test/java/features/Dropdown Tests.feature
```diff
@@ -7,31 +7,31 @@ Feature: Dropdown Tests
   Scenario: 273 Dropdown
     Given I am on the Dropdown Page
     When I select Option 1 from the Dropdown
-      And I verify the Dropdown has the text "Option 1" selected
+      And I verify the Dropdown has the text "Option 1"
```
dougnoel, line 10: If we added `(?:.*)` to the end of line 151 in TextVerificationSteps, the wording of this step wouldn't have to change and the quotes would make more sense.

dougnoel (approved):

dougnoel (dismissed):

dougnoel (approved):

PR #462, opened 2022-12-08
Api

dougnoel:
Adds in API testing for GET, POST, PUT and DELETE.


src/main/java/com/dougnoel/sentinel/apis/API.java
```diff
@@ -0,0 +1,57 @@
...
+    public String getName() {
+        return apiName;
+    }
+		
+	/**
+	 * Returns a java.net.URI constructed from the URL listed in the API yaml file.
+	 * 
+	 * @return java.net.URI the constructed URI
+	 */
+	protected URIBuilder getURIBuilder(String passedText) {
```
sampacos, line 40: weird indenting here

src/main/java/com/dougnoel/sentinel/steps/APISteps.java
```diff
@@ -0,0 +1,173 @@
...
+	public void setAPI(String apiName) {
+        APIManager.setAPI(apiName);
+	}
+
+	/**
+	 * Sets the body of the active API call to the string passed.
+	 * 
+	 * @param body String the json to be passed as the body of the request.
+	 */
+	@Given("I set the API body to")
```
sampacos, line 45: Missing capture group, no?

sampacos: Also, does it make more sense to say we are setting the "request body", rather than the "API body"?

src/test/java/features/89 API Testing.feature
```diff
@@ -0,0 +1,65 @@
+#language: en
+@89
+Feature: 89 API Testing
+  Tests using the Swagger Pet Store example API.
+  https://petstore3.swagger.io/
+      
+  @89A
+  Scenario: 89A POST Swagger Test
+    Given I use the API named Pet Store API
+    When I load puppydata to use as the API body
```
sampacos, line 10: same thing here. To me, the phrase "API body" doesn't really make sense

src/main/java/com/dougnoel/sentinel/apis/API.java
dougnoel: That is weird. It doesn't look that way in the IDE.

src/main/java/com/dougnoel/sentinel/enums/Type.java
```diff
@@ -0,0 +1,13 @@
+package com.dougnoel.sentinel.enums;
+
+/**
+ * Possible methods for an API to authenticate.
+ * 
+ * @author Doug Noël
+ *
+ */
+public enum Type {
```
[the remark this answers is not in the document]
dougnoel: It was AuthenticationType. I thought I had deleted the file completely. Somehow...not? It is now deleted until we actually implement it.

src/main/java/com/dougnoel/sentinel/steps/APISteps.java
dougnoel: Good call in the naming. Fixed another step as well. We aren't missing a capture group. You'll notice this isn't surrounded by a ^$. We are actually using this as a cucumber step that doesn't use regex, and so it expects an argument through some magic I do not understand. It works however.

src/test/java/features/89 API Testing.feature
dougnoel: Fixed, as well as in the load request body step.

pturchinetz (approved):

sampacos (approved):

PR #474, opened 2022-12-19
468 more csv

sampacos:
Add more csv capabilities. Add ability to check raw text (UTF-8) in any downloaded file.


dougnoel (approved):
We should look at an append method for writing as writing the whole file every time could get expensive but we should see how it's used.

PR #475, opened 2022-12-20
Api

dougnoel:
Adds variables and the ability to test response time.


sampacos (dismissed):

src/main/java/com/dougnoel/sentinel/system/FileManager.java
```diff
@@ -212,43 +207,6 @@ public static String sanitizeString(String toSanitize) {
...
-	 * <b>Supported Path Shortcuts:</b>
-	 * <ul>
-	 * <li>%localappdata% <i>- Ex: C:\Users\USERNAME\AppData\Local</i></li>
-	 * <li>%appdata% <i>- Ex: C:\Users\USERNAME\AppData\Roaming</i></li>
-	 * <li>%USERPROFILE% <i>- Ex: C:\Users\USERNAME</i></li>
-	 * </ul>
-	 * @param pathToProcess String path to replace windows special folder shortcuts
-	 * @return String the path string with windows special folder shortcuts replaced with their environment equivalent path
-	 */
-	public static String winSpecialFolderConverter(String pathToProcess){
```
sampacos, line 227: Are we sure we don't need this for curation tools testing? I have a feeling this was added for a reason

dougnoel: This was only called by WinAppDriver creation, and we already handled this issue in the loading of the executable from the page object. All the Curation tests which are screenshot based, use a simpler one-liner. I removed this because this method and its attendant 4 unit tests were causing intermittent failures in the build pipeline. If we do find that the code using it needs to be updated I'll show Turch how to update it.

dougnoel: I'll make sure path variables work before removing this.

PR #484, opened 2023-02-01
Windows Application Testing refactor to Appium 1.2

pturchinetz:
Changes the Sentinel setup to use Appium 1.2 as the main handler for WindowsDriver instead of previously handling it ourselves. This should streamline the debugging and reduce weirdness in the code from our previous implementation.


src/main/java/com/dougnoel/sentinel/webdrivers/WindowsDriverFactory.java
```diff
@@ -0,0 +1,90 @@
...
+	/**
+	 * Exists to defeat instantiation.
+	 */
+	private WindowsDriverFactory() {}
+
+	protected static void startAppiumService() {
+		AppiumServiceBuilder builder = new AppiumServiceBuilder();
+		builder.withIPAddress("127.0.0.1").withArgument(BASEPATH , "/wd/hub").usingPort(4725);
+		appiumService = AppiumDriverLocalService.buildService(builder);
+		appiumService.isRunning();
```
sampacos, line 35: Will this throw an exception if the service isn't running? If not, and it returns a boolean, should we do something with the return value?

sampacos (approved):
great work! just the one question but nothing blocking imo

src/main/java/com/dougnoel/sentinel/webdrivers/WindowsDriverFactory.java
pturchinetz: If something goes wrong and the service fails to start (or the isRunning check fails) it fails out and stops, which I think is the behavior we want? I'm really using isRunning like an implicit wait, since earlier on the tests would start before the Appium server was properly spun up.

PR #486, opened 2023-02-01
Tyler/documentation fix minor api changes

tyBouch:
Adds some storage and comparison of a response body against stored values.
Fixes some missing documentation in two locations.
Changed a step as using quotations was preventing proper checking of values (values being compared often contain quotations themselves).
Adjusted tests accordingly.


dougnoel:
Closed. See the new pull request I just created.

PR #488, opened 2023-02-03
Updated code for the way API Response objects are stored and compared.

dougnoel:


dougnoel:
Sceanrios need to store all the responses. After a scenario is over, we need to clear it.

Issue #491, opened 2023-02-09
Table verification: need ability to numerically compare values in a column to a reference value


sampacos:
For a table such as
Column A | Column B
X | 1.2
Y | 3.1
Z | 0.1

I need to be able to verify that all values in Column B are less than a certain value (or greater than, or equal to).

PR #492, opened 2023-02-09
Add table verification step. Add screenshots for failing tests

sampacos:
- Address #491, adding verification step for numeric values in table column.
- Add hooks file to take a screenshot (only of the application window) if the test fails and attach it to the report.

### DIFF
--- a/src/main/java/com/dougnoel/sentinel/elements/tables/Table.java
+++ b/src/main/java/com/dougnoel/sentinel/elements/tables/Table.java
@@ -1,11 +1,13 @@
 package com.dougnoel.sentinel.elements.tables;
 
+import java.math.BigDecimal;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.stream.Collectors;
 
 import com.dougnoel.sentinel.steps.BaseSteps;
 import com.dougnoel.sentinel.webdrivers.WebDriverFactory;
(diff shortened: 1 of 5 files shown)


src/main/java/com/dougnoel/sentinel/elements/tables/Table.java
```diff
@@ -973,4 +974,23 @@ public boolean waitForSpecificCellToContain(int numberOfSecondsToWait, String co
...
+	 * Converts all values in the given column to double.
+	 * Valid comparisonType values = {"less than", "greater than"}. Any other value of comparisonType will make this method perform an "equals" comparison.
+	 * @param columnHeader String name of the column
+	 * @param comparisonType String type of comparison to perform. Valid values are "less than", "greater than". Any other string will force an "equals" comparison.
+	 * @param referenceNumber double the number to compare against.
+	 * @return boolean true if all values in the column satisfy the given comparison. false otherwise.
+	 */
+	public boolean verifyNumericValuesInWholeColumn(String columnHeader, String comparisonType, double referenceNumber){
+		var allColumnData = getAllCellDataForColumn(columnHeader).stream().map(Double::parseDouble).collect(Collectors.toList());
+		if(comparisonType.equalsIgnoreCase("less than"))
```
dougnoel, line 989: Do this as a case statement. You can filter the comparisonType through to make it all lowercase before the comparison. This will be easier to read and easier for someone else to modify later if they want to add stuff like greater than or equal to or not equal to.

dougnoel (changes requested):

dougnoel (approved):

PR #501, opened 2023-02-17
add support to use stored variables in APISteps params and header

david-ondrus:
add support to use stored variables in request body


src/main/java/com/dougnoel/sentinel/strings/SentinelStringUtils.java
```diff
@@ -132,4 +132,54 @@ public static String replaceVariable(String text) {
...
+				text = StringUtils.replaceOnce(text, variable, value); //Using replaceOnce so that if we have the same variable name twice we do not run into iteration issues.
+			}
+			else{
+				text = StringUtils.replaceOnce(text, variable, "\""+value+"\""); //Using replaceOnce so that if we have the same variable name twice we do not run into iteration issues.
+
+			}
+			}
+		return text;
+	}
+	public static String replaceURLVariable(String text) {
```
pturchinetz, line 175: needs header

src/main/java/com/dougnoel/sentinel/strings/SentinelStringUtils.java
```diff
@@ -132,4 +132,62 @@ public static String replaceVariable(String text) {
...
+	 * 	 "name": "puppy",
+	 * 	 "category": {
+	 * 	 "id": 1,
+	 * 	 "name": "Dog"
+	 *    }}.
+	 * If there are no values to replace, this method returns the string intact.
+	 * @param text String the text to search for variable replacement
+	 * @return String the string with variables replaced as applicable
+	 */
+	public static String replaceStoredVariables(String text) {
```
sampacos, line 157: If I'm honest I don't like the design of the replacement syntax. If you need to surround something with quotes, why not just have the quotes in the template? For example, if you need the result to look like `"name": "Dog"`, why not make the input `"name": "{category_name}"`. It seems to me that's the only distinction made between `int` and `str` here, but I think it makes more sense to not have any such distinction. If there is a question of whether `{variable_name}` is unique enough, then we can change the syntax to something more unique without adding type info to the front.

sampacos (changes requested):

src/main/java/com/dougnoel/sentinel/steps/APISteps.java
```diff
@@ -99,7 +99,8 @@ public void addParameter(String parameter, String value) {
 	 */
 	@When("^I send a (DELETE|GET|POST|PUT) request to the (.*?) endpoint$")
 	public void sendRequest(String apiCallType, String endpoint) {
-		APIManager.sendRequest(RequestType.valueOf(apiCallType), endpoint);
+		var endpointFormatted = SentinelStringUtils.replaceStoredVariables(endpoint);
```
dougnoel, line 102: This should be inline to be consistent with other usages.

src/main/java/com/dougnoel/sentinel/steps/APISteps.java
```diff
@@ -189,7 +190,17 @@ public void verifyResponseContains(String assertion, String matchType, String te
...
-		APIManager.addHeader(name, value);
+		APIManager.addHeader(name, SentinelStringUtils.replaceStoredVariables(value));
 	}
 
+	/**
+	 * Sets the body of the active API call to the string passed.
+	 *
+	 * @param body String the json to be passed as the body of the request.
+	 */
+	@Given("I set the request body with variables to")
```
dougnoel, line 201: I don't know what this step means. The steps need to read like manual testing steps. If this test were to fail and I were to have to manually reproduce this, what would the step look like to reproduce what's happening here? There are also no usage examples in the Javadoc above. If this is an internal helper method, there should be no Given decorator, and this method should be private.

src/test/java/features/89 API Testing.feature
```diff
@@ -66,10 +66,55 @@ Feature: 89 API Testing
...
     Then I verify the response code equals 200
     When I GET record 10 from the pet endpoint
-    Then I verify the response code equals 404
\ No newline at end of file
+    Then I verify the response code equals 404
+
+  @89G
+  Scenario: 89G Body With Parameters Test
+    Given I use the API named Pet Store API
+    When I initialize the data
```
dougnoel, line 80: When I initialize the configuration values as follows
"""
id: 10
category_name: puppies
"""

dougnoel (dismissed):

pturchinetz (approved):

PR #512, opened 2023-03-03
Tyler/508 filename date time verification

tyBouch:


src/main/java/com/dougnoel/sentinel/steps/DownloadVerificationSteps.java
```diff
@@ -61,6 +63,34 @@ public static void verifyFileWithFilenameDownloaded(String expectedFilename) thr
...
+	 */
+	@Then("^I verify the filename contains today's date(?: formatted as )?(.*?)$")
+	public static void verifyFileWithDateTimeFilename(String dateTimePattern) {
+		if(dateTimePattern.isBlank())
+			dateTimePattern = "YYYY_MMM_dd";
+
+		DateTimeFormatter dateFormatter = DateTimeFormatter.ofPattern(dateTimePattern);
+		String formattedDateTime = dateFormatter.format(LocalDateTime.now());
+
+		String mostRecentDownloadPath = DownloadManager.getMostRecentDownloadPath().toString();
```
sampacos, line 86: this path is the fully qualified path on the system. for example, "C:/source/sentinel/downloads/newdownload.pdf". You should grab just the filename from the path for the comparison, since that's what the step says it does.

sampacos (changes requested):

src/main/java/com/dougnoel/sentinel/steps/DownloadVerificationSteps.java
tyBouch: Aye I'll make the change!

sampacos (approved):

PR #519, opened 2023-03-08
Added comments specifying the github pages for more information on se…

tyBouch:
…tting configurations utilizing the yml configuration method, and that the example.sentinel.yml is not used.


conf/example.sentinel.yml
```diff
@@ -1,4 +1,7 @@
 ---
+#This file is provided as an example, and is not used as-is.
+#Please create a "sentinel.yml" file to set configurations using this method.
```
pturchinetz, line 3: I'm 99% sure this is correct but I want @dougnoel to verify this

dougnoel (approved):

PR #521, opened 2023-03-08
Removed the required quotations

tyBouch:
Updated documentation with gherkin examples
Updated the feature file testing API calls to reflect the change.

### DIFF
--- a/src/main/java/com/dougnoel/sentinel/steps/APISteps.java
+++ b/src/main/java/com/dougnoel/sentinel/steps/APISteps.java
@@ -94,9 +94,9 @@ public class APISteps {
      * <li>I send a PUT request to the amdins endpoint</li>
      * </ul>
      * <p>
-     *  
-	 * @param apiCallType
-	 * @param endpoint
+     *
+	 * @param apiCallType String the type of call to make
+	 * @param endpoint String the endpoint name as referenced in the swagger file
 	 */
 	@When("^I send a (DELETE|GET|POST|PUT) request to the (.*?) endpoint$")
 	public static void sendRequest(String apiCallType, String endpoint) {
@@ -150,12 +150,19 @@ public class APISteps {
 
 	/**
 	 * Validates text in an API response.
-	 * 
+	 * <p>
+	 * <b>Gherkin Examples:</b>
+	 * <ul>
+	 * <li>I validate the response <b>contains</b> the text <b>"key":"value"</b></li>
+	 * <li>I validate the response <b>does not contain</b> the text <b>"value":"key"</b></li>
+	 * <li>I validate the response <b>has</b> the text <b>123456789ABCDEFG</b></li>
+	 * </ul>
+	 * <p>
 	 * @param assertion String null to see if the text exists, "does not" to see if it is absent
 	 * @param matchType String use "contains" for a partial match otherwise it will be an exact match
 	 * @param text String the text to match
 	 */
-	@Then("^I validate the response( does not)? (has|have|contains?) the text \"([^\"]*)\"$")
+	@Then("^I validate the response( does not)? (has|have|contains?) the text (.*?)$")
     public static void verifyResponseContains(String assertion, String matchType, String text) {
         boolean negate = !StringUtils.isEmpty(assertion);
         boolean partialMatch = matchType.contains("contain");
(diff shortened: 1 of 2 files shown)


pturchinetz (dismissed):

src/main/java/com/dougnoel/sentinel/steps/APISteps.java
```diff
@@ -150,12 +150,19 @@ public static void verifyResponseTime(double time) {
...
+	 * <li>I validate the response <b>does not contain</b> the text <b>"value":"key"</b></li>
+	 * <li>I validate the response <b>has</b> the text <b>123456789ABCDEFG</b></li>
+	 * </ul>
+	 * <p>
 	 * @param assertion String null to see if the text exists, "does not" to see if it is absent
 	 * @param matchType String use "contains" for a partial match otherwise it will be an exact match
 	 * @param text String the text to match
 	 */
-	@Then("^I validate the response( does not)? (has|have|contains?) the text \"([^\"]*)\"$")
+	@Then("^I validate the response( does not)? (has|have|contains?) the text (.*?)$")
```
dougnoel, line 165: The way to solve this si to write our own contains and equals evaluation methods in SentinelStringUtils that checks the conditions with surrounding quotes, and then without before returning a result so that we don't fail based on someones assumptions and so we don't break old code.

dougnoel (changes requested):

src/main/java/com/dougnoel/sentinel/steps/APISteps.java
pturchinetz: I know we had similar discussion for text verification but it looks like we never implemented those changes (and probably should)?
TextVerificationSteps line 74: `@Then("^I verify the (.*?)( does not)? (has|have|contains?) the text \"([^\"]*)\"$")`
Last I checked we wanted to drop quote checking to free the user to put in exactly what they wanted, lest we run slower to accommodate them. I agree that we don't want to break old code but we're rapidly approaching a 2.0 if we're slowing down our code like this.

src/main/java/com/dougnoel/sentinel/steps/APISteps.java
```diff
@@ -150,37 +150,65 @@ public static void verifyResponseTime(double time) {
...
+		boolean negateResult;
+		boolean result;
+		boolean textStartEndQuotes = false;
+		String startEndQuoteStripText = text;
+		if(text.matches("^\".*?\"$")) {
+			textStartEndQuotes = true;
+			startEndQuoteStripText = text.substring(1, text.length() - 1);
+		}
+
+		if (partialMatch) {
```
pturchinetz, line 187: To reduce code complexity you'll need to break this out into a different function, and at that point it should also be applied to TextVerificationSteps.java to have them follow this new pattern.
Sorry to increase your scope!

src/main/java/com/dougnoel/sentinel/steps/APISteps.java
```diff
@@ -150,37 +150,65 @@ public static void verifyResponseTime(double time) {
...
+			else {
+				result = responseText.contains(text);
+				negateResult = result;
+			}
+		}
+
+		if (negate)
+			assertFalse(expectedResult, negateResult);
+		else
+			assertTrue(expectedResult, result);
```
dougnoel, line 210: We had a miscommunictaion. This is what I was attempting to describe:

```suggestion
if (partialMatch) {
            if (negate) {
                assertTrue(expectedResult, SentinelStringUtils.doesNotContain(responseText, text));
            } else {
                assertTrue(expectedResult, SentinelStringUtils.contains(responseText, text));
            }
        } else {
            if (negate) {
                assertTrue(expectedResult, SentinelStringUtils.doesNotequal(responseText, text));
            } else {
                assertTrue(expectedResult, SentinelStringUtils.equals(responseText, text));
            }
        }
```

Those four methods need to be created in SentinelStringUtils, based on our conversations. Then we can reuse those methods in our other text verification areas.

dougnoel (changes requested):

src/main/java/com/dougnoel/sentinel/steps/APISteps.java
tyBouch: Same logic at least, but yes Turch and I came to this conclusion. Though where to put it we weren't entirely sure so this helps a lot.

PR #526, opened 2023-03-13
525 specific cell value set delete entire row

tyBouch:


tyBouch:
Will update the unit tests soon.

dougnoel (approved):

PR #552, opened 2023-06-02
551 zip file count verification

tyBouch:
Adds a new step which can take an extension optionally, and a file count.

File count can be any integer.
"0" is valid as an optional way of checking for 0 files. Otherwise "any" can be utilized for "above 0" or none for does not.
Uses file content stream filtering to filter down the extensions, or gets the flat file names size otherwise.

Only takes 0+, only whole numbers or "any" for the count of files. Any extension is allowed, but it must end with the given extension.

Old step "^I verify the most recently downloaded zip file( does not)? contains? a file with the extension (.*?)$" has been updated to use the new step with the "any" call, and has had its documentation updated that the steps are equivalent and will behave the same.

Also supports case-insensitive extension or case sensitive partial name contains. Good for testing for partially randomly generated names like reports with a date-time stamp. Extension is case insensitive as the OS doesn't care about casing. Naming is case sensitive.

### DIFF
--- a/src/test/java/features/327 Download Verification Steps.feature
+++ b/src/test/java/features/327 Download Verification Steps.feature
@@ -40,21 +40,21 @@ Feature: Download Verification Steps
   @551A
   Scenario: Download zip and verify it contains 1 pdf
     Then I verify that by clicking the zip download link a new file is downloaded with the extension zip
-      And I verify the most recently downloaded zip file contains 1 file with the extension pdf
+      And I verify the most recently downloaded zip file contains 1 file with extension pdf
 
   @551B
   Scenario: Download zip and verify it contains no files with png
     Then I verify that by clicking the zip download link a new file is downloaded with the extension zip
-      And I verify the most recently downloaded zip file contains 0 files with the extension png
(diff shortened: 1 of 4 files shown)


src/test/java/features/327 Download Verification Steps.feature
```diff
@@ -35,4 +35,50 @@ Feature: Download Verification Steps
...
+    Then I verify that by clicking the zip download link a new file is downloaded with the extension zip
+      And I verify the most recently downloaded zip file contains 0 files with extension png
+      And I verify the most recently downloaded zip file does not contains 1 file with extension png
+      And I verify the most recently downloaded zip file does not contain any files with extension png
+
+  @551C
+  Scenario: Download zip and verify it does not contain 5 pdfs, contains 1 pdf, and contains 1 html
+    Then I verify that by clicking the zip download link a new file is downloaded with the extension zip
+    And I verify the most recently downloaded zip file does not contain 5 files with extension pdf
+    And I verify the most recently downloaded zip file contains 1 file with extension pdF
```
pturchinetz, line 56: clever catch

pturchinetz (approved):

